PR #36, opened 2020-03-25
Add support for setting HOST_UID and HOST_GID from environment

jswelling:


yuanzhou:
Thanks @jswelling, question: in your improvement, the HOST_UID and HOST_GID will be grabbed from the variables that gateway `hubmap-docker.sh` exported. There are a lot of times that I'll need to re-deploy one of the APIs. Will those exported  HOST_UID and HOST_GID still be available so it doesn't fall back to the default 1000?

jswelling:
hubmap_docker.sh only sets HOST_UID and _GID from the environment in the localhost case, which I *think* is always correct.  For all other cases HOST_UID and _GID are set from environment variables if those are present, and default to the given values of 1000 otherwise.  I don't know the mechanics of how deployments like dev are set up, but if the environment variables are not set things should fall back to their previous default values.

PR #54, opened 2020-06-24
Entity access level zhou

yuanzhou:
- added new API endpoint `/entity-access-level/<uuid>` to return the access level as string for a given entity
- this endpoint calls the `dataset. get_entity_access_level()`
- improved error handling
- also added this new endpoint in gateway (in a separate branch now) and the auth requires HuBMAP-Read group.

@shirey @cborromeo please review and merge if all looks good. I've tested and it works with the gateway on DEV. One issue I noticed is that in the DEV version of neo4j, many of the entity nodes don't have the `data_access_level` attribute at this point since it's newly added.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -188,6 +188,11 @@ def get_entity_access_level(uuid):
         logger.error(msg, exc_info=True)
         print(msg)
         return Response(msg, 500)
+    except Exception as e:
+        msg = 'Unhandled exception occured, check log file for detail'
+        print (msg)
+        logger.error(msg, exc_info=True)
+        return Response(msg, 500)
 
 
 @app.route('/entities/uuid/<uuid>', methods = ['GET'])
--- a/src/app.py
+++ b/src/app.py
@@ -180,13 +180,12 @@ def get_entity_access_level(uuid):
         return dataset.get_entity_access_level(uuid)
     except HTTPException as hte:
         msg = "HTTPException during get_entity_access_level HTTP code: " + str(hte.get_status_code()) + " " + hte.get_description() 
-        print(msg)
         logger.warn(msg, exc_info=True)
         return Response(hte.get_description(), hte.get_status_code())
     except CypherError as ce:
-        print ('A Cypher error was encountered when calling dataset.get_entity_access_level(): ' + ce.message)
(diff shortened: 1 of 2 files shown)


yuanzhou:
> This looks good. Just make sure my modifications to the entity-access-level-zhou code make it into master

Thanks!

src/app.py
```diff
@@ -170,6 +170,26 @@ def get_entity(identifier):
...
+    except HTTPException as hte:
+        msg = "HTTPException during get_entity_access_level HTTP code: " + str(hte.get_status_code()) + " " + hte.get_description() 
+        logger.warn(msg, exc_info=True)
+        print(msg)
+        return Response(hte.get_description(), hte.get_status_code())
+    except CypherError as ce:
+        msg = 'A Cypher error was encountered when calling dataset.get_entity_access_level(), check log file for detail'
+        logger.error(msg, exc_info=True)
+        print(msg)
+        return Response(msg, 500)
```
shirey, line 190: This should catch a general Exception here instead of just a CypherError, that way if unexpected errors occur we'll cleanly handle

src/dataset.py
```diff
@@ -254,26 +254,29 @@ def get_activity_output_for_dataset(self, driver, uuid):
...
+                raise HTTPException("Entity uuid:" + uuid + " not found.", 404)
             if len(return_list) > 1:
-                raise HTTPException("Multiple datasets found for uuid:" + uuid, 500)
+                raise HTTPException("Multiple entities found for uuid: " + uuid, 500)
             return return_list[0]
                         
-        except Exception as e:
-            raise HTTPException("Unhandled Exception occurred", 500)
+        except CypherError as ce:
+            raise CypherError('A Cypher error was encountered: ' + ce.message)
```
shirey, line 279: Same issue as above, we need to cleanly handle any exception that occurs, not just CypherErrors.  Unforseen, but uncommon, exceptions could occur in code (python or third-party libraries) that we don't control.

yuanzhou:
@shirey thanks for the code review, I'll add the general Exception handling.

yuanzhou:
@shirey just added the general Exception handling. Thanks for pointing that out.

PR #62, opened 2020-08-04
made both collections calls available for consortium/public data without an auth token

shirey:
@cborromeo please give a simple review
@yuanzhou  please deploy to DEV when Chuck gives the okay


yuanzhou:
Currently gateway api lookup doesn’t require token on “/collections” but a HuBMAP-read token required for “/collections/<id>”. Do I need to disable the auth check?

shirey:
@yuanzhou Yes, remove the requirement on /collections/ too as a token is no longer required, but if there is one I check it inside the method.

PR #131, opened 2021-03-16
added DataSubmission entity to schema.yaml

shirey:


src/schema/provenance_schema.yaml
```diff
@@ -576,5 +584,69 @@ ENTITIES:
         # The updated_peripherally tag is a temporary measure to correctly handle any attributes
         # which are potentially updated by multiple triggers
         updated_peripherally: true
-      
-  
\ No newline at end of file
+
+  ############################################# Sample #############################################
+  DataSubmission:
+    # TODO: Zhou, not sure what these derivation properties should be set to.  DataSubmission
```
yuanzhou, line 590: @shirey I introduced this `derivation` property to indicate if the entity class/type can be used as source of another entity class/type and on the other hand if this entity class/type can be derived from another class/type. For instance, a Dataset can be created from another Dataset/Sample, but can't derived from Donor/Collection. This rule will be used mainly by validations to prevent end users from abusing the system.

yuanzhou:
I'll merge this PR and add additional functionalities to the `feature/data-submission` branch.

src/schema/provenance_schema.yaml
shirey: So, @yuanzhou, should source be true in this case and target be false because the ancestor needs to be a Lab?  Or is the connection to Lab not considered here because it is via the special group_uuid mechanism?

PR #149, opened 2021-04-02
Feature/data submission

yuanzhou:
- Add Submission definition to yaml schema
- Add trigger methods used by Submission properties
- Tweaks to `create_hubmap_ids()` to require group id for Submission and error handlings
- Neo4j cypher queries to work with the Submission linkages

The above changes have been merged into `dev-integrate` and tested on DEV along with the search-api changes.


src/app.py
```diff
@@ -2294,6 +2329,16 @@ def require_json(request):
     if not request.is_json:
         bad_request_error("A json body and appropriate Content-Type header are required")
 
+"""
+Check if the token from user request is valid
+
+token : str or Response
+    The token string if valid or Flask Response object if invalid
+"""
+def require_token(token):
```
shirey, line 2338: By doing this you lose the information returned in that response.  If you return the response then the custom message set within and specific code (400 vs 401 vs 403, but maybe sometimes 500) is sent back to the requester.

shirey (changes requested):
@yuanzhou see my single comment below.  I don't think it is good to swallow that information...  If you want to centralize the handling of these the general response exception (msg and status code) could be used by grabbing the message and status code from the response and throwing this exception.

src/app.py
yuanzhou: Thanks @shirey , that's a good point. I'll make the changes.

yuanzhou:
Made the change with returning the original message:
````
def require_token(token):
    if isinstance(token, Response):
        # When the token is an Response instance,
        # it must be a 401 error with message
        # We wrap the message in a json and send back to requester as 401 too
        # The Response.data returns binary string, need to decode
        unauthorized_error(token.data.decode())
````

The https://github.com/hubmapconsortium/commons/blob/test-release/hubmap_commons/hm_auth.py#L194 only returns as 401 when the result is a `Response` object other than the token string. That's why I only used 
````
unauthorized_error(token.data.decode())
````

yuanzhou:
And I also made the `/entities/<id>` endpoint not to require a token in gateway: https://github.com/hubmapconsortium/gateway/commit/51aacdc51b52e2e7c1985f7dc4682c046b60dc06 so entity-api will be able to response with the information of public entity.

shirey (approved):
@yuanzhou This works slightly better, but there is a chance that another code, such as a 500 is thrown. Fine to use this in this instance and I think sending a Response object as a reply from commons is definitely not the best way to do this.

Let's get something in place in the future that can generally handle errors by responding with user friendly error messages and correct http status codes.

yuanzhou:
@shirey you are right. I've been thinking about the error handling too. I'll get this out first then come back with the error handling, it'll require changes in several places.

PR #170, opened 2021-05-11
Improving documentation

shirey:
- added framework to host docs on GitHuB pages
 - added schema information, request and response details to openapi spec/SmartAPI docs


entity-api-spec.yaml
```diff
@@ -1,396 +1,1267 @@
...
+            - protein
+            - ran_poly_a_enriched
+            - rna_total
+            - rnalater_treated_and_stored
+            - rnaseq
+            - scatacseq
+            - scrnaseq
+            - segment
+            - seqfish
+            - sequence_library
```
yuanzhou, line 312: `sequence_library` appeared twice in the list here. 

 I also noticed that we have two entries of `sequence_library` in the yaml definitions:
- https://github.com/hubmapconsortium/search-api/blob/test-release/src/search-schema/data/definitions/enums/tissue_sample_types.yaml#L79
- https://github.com/hubmapconsortium/search-api/blob/test-release/src/search-schema/data/definitions/enums/tissue_sample_types.yaml#L83

Their descriptions are slightly different with one has lowercased library and the other capitalized.

entity-api-spec.yaml
```diff
@@ -1,396 +1,1267 @@
...
-        "404":
-          description: The target entity could not be found
-        "500":
+        '500':
           description: Internal error
-          
-  /entities/dataset/globus-url/{id}:
-    put:
-      summary: 'Get the Globus URL to the given dataset'
+  '/entities/dataset/globus-url/{id}':
```
yuanzhou, line 1210: The current entity-api also supports the old endpoint `/entities/dataset/globus-url/<id>`

https://github.com/hubmapconsortium/entity-api/blob/test-release/src/app.py#L1733

Should we add it to the API docs too?

entity-api-spec.yaml
```diff
@@ -1,396 +1,1267 @@
...
+        '404':
           description: The target entity could not be found
-        "500":
+        '500':
           description: Internal error
-          
-  /collection/redirect/{id}:
-    put:
-      summary: 'Redirect a request from a doi service for a collection of data'
+  '/doi/redirect/{id}':
```
yuanzhou, line 1193: The current entity-api also supports the old endpoint `/collection/redirect/<id>`

https://github.com/hubmapconsortium/entity-api/blob/test-release/src/app.py#L1644

Should we add it to the API docs too?

entity-api-spec.yaml
shirey: No, let's leave this one out of the docs.  The old one will be "deprecated forever".

entity-api-spec.yaml
shirey: Same thing with this one, I purposely left it out of the documentation because it is "deprecated forever".

entity-api-spec.yaml
```diff
@@ -1,396 +1,1267 @@
...
+      summary: 'Returns the information of the Collection specified by the uuid with all connected datasets.  For this call, the JWT is optional. If a valid token is provided with group membership in the HuBMAP-Read group any collection matching the id will be returned.  Otherwise if no token is provided or a valid token with no HuBMAP-Read group membership then only a public collection will be returned.  Public collections are defined as being published via a DOI (collection.doi_registered == true) and at least one of the connected datasets is public (dataset.metadata.data_access_level == ''public''). For public collections only connected datasets that are public are returned with it.'
```
yuanzhou, line 1073: It says the `JWT`, are we using JWT token at all? Prob a typo.

shirey: Yes, JWT doesn't belong there.  Looks like it was there before and I didn't remove it.

yuanzhou (approved):

PR #207, opened 2021-07-19
reorganized schema template work undir its own directory

shirey:
Moved all of the schema templating code into its own directory.


DerekFurstPitt (approved):
Trying out pull request the workflow

yuanzhou (approved):
Looks good

DerekFurstPitt:
@shirey checked out locally to verify everything still worked. A couple file paths were hard coded so they need fixed. i'll fix these form my end. No action needed from your end

yuanzhou:
@DerekFurstPitt just a bit more on the Pull Request (PR) workflow. Once a PR is merged you can:
- delete the branch that made the changes if you are sure it won't be reused by the branch creator, especially when you are the branch creator.
- keep it around until the branch creator agrees it can be deleted.

Github allows us to automatically delete the merged branch with settings, we don't set that way because often times 
- you may just continue make tweaks to a merged branch
- your branch gets merged into `dev-integrate` for DEV testing, then the same branch will need get into `test-release` after DEV testing.

DerekFurstPitt:
@yuanzhou Yep, I double checked with bill, its was ok to delete his branch. thanks for the info

PR #227, opened 2021-09-22
Derek furst/datasets feature retract

DerekFurstPitt:
Re-issuing pull request. Accidentally issued pull request from my feature branch directly into test-release instead of dev-integrate. Added /datasets/id/revisions. Also added proper authorization. Only had a small amount of data to test with (dev neo4j only has a handful of versioned dataset) but all of them passed my tests and performed as expected. I also applied the latest changes from test-release.


yuanzhou:
I added more validations at the schema level to prevent the users from abusing the generic entity update call to update the retraction stuff.

yuanzhou (approved):

PR #229, opened 2021-09-23
Separating revisions and retract methods into 2 separate branches. Th…

DerekFurstPitt:
Separated datasets/<id>/revisions and datasets/<id>/retract into 2 separate feature branches. Retract is still a work in progress (PR changed to draft). This one passed all local testing and is ready to be tested on dev-integrate assuming no merge conflicts or requested changes.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1955,6 +1955,8 @@ def get_dataset_revision_number(id):
     # Response with the integer
     return jsonify(revision_number)
 
+
+
 @app.route('/datasets/<id>/revisions', methods=['GET'])
 def get_revisions_list(id):
     # Token is not required, but if an invalid token provided,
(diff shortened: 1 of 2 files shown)


src/app.py
```diff
@@ -1953,6 +1955,73 @@ def get_dataset_revision_number(id):
...
+            user_token = auth_helper_instance.getAuthorizationTokens(request.headers)
+            if not user_in_hubmap_read_group(request) or isinstance(user_token, Response):
+                can_display = False
+                if 'next_revision_uuid' in outputjson[-1]['dataset']:
+                    outputjson[-1]['dataset'].pop('next_revision_uuid')
+        if can_display:
+            outputjson.append(revision)
+        total_revisions = total_revisions + 1
+    outputjson.reverse()
+    return Response(json.dumps(outputjson, sort_keys=True), 200, mimetype='application/json')
```
yuanzhou, line 2024: Better to use Flask's `jsonify()` which can serialize python list/dict to JSON and wrap it in a Response with the application/json mimetype.

src/app.py
```diff
@@ -1953,6 +1955,73 @@ def get_dataset_revision_number(id):
     # Response with the integer
     return jsonify(revision_number)
 
+@app.route('/datasets/<id>/revisions', methods=['GET'])
```
yuanzhou, line 1958: To be consistent, please add a comment block that describes this call, the parameters, and return type above this line. Follow examples from other endpoints in this file.

src/app.py
```diff
@@ -1953,6 +1955,73 @@ def get_dataset_revision_number(id):
...
+    for j in complete_next_list:
+        can_display = True
+        revision = collections.OrderedDict()
+        j = schema_manager.normalize_entity_result_for_response(j)
+        revision['dataset'] = j
+        revision['dataset_uuid'] = j['uuid']
+        revision['revision_number'] = total_revisions
+        if j['status'].lower() != DATASET_STATUS_PUBLISHED:
+            user_token = auth_helper_instance.getAuthorizationTokens(request.headers)
+            if not user_in_hubmap_read_group(request) or isinstance(user_token, Response):
```
yuanzhou, line 2016: Need to rework this logic for handling the very last revision based on its status and the user token.

src/app.py
```diff
@@ -1953,6 +1955,73 @@ def get_dataset_revision_number(id):
...
+    final_result = schema_manager.normalize_entity_result_for_response(complete_dict)
+    uuid = final_result['uuid']
+    previous_revisions = app_neo4j_queries.get_previous_revisions(neo4j_driver_instance, uuid)
+    complete_previous_list = schema_manager.get_complete_entities_list(token, previous_revisions)
+    next_revisions = app_neo4j_queries.get_next_revisions(neo4j_driver_instance, uuid)
+    complete_next_list = schema_manager.get_complete_entities_list(token, next_revisions)
+    outputjson = []
+    total_revisions = 1
+    for i in complete_previous_list:
+        revision = collections.OrderedDict()
```
yuanzhou, line 1994: I don't see the need to use this `collections.OrderedDict()` because we don't care about the ordering of these three properties: `dataset`, `dataset_uuid`, and `revision_number`. And they should be consistent without enforcing ordering from the result anyway.

src/app.py
yuanzhou, line 1958: To be consistent, please add a comment block that describes this call, the parameters, and return type above this line. Follow examples from other endpoints in this file.

src/app.py
yuanzhou, line 1994: I don't see the need to use this `collections.OrderedDict()` because we don't care about the ordering of these three properties: `dataset`, `dataset_uuid`, and `revision_number`. And they should be consistent without enforcing ordering from the result anyway.

src/app.py
yuanzhou, line 2016: Need to rework this logic for handling the very last revision based on its status and the user token. Your code currently skips this next revision dataset if it's non-published and the user doesn't has access (no token or not in the HuBMAP-Read group). This would work correctly only when this dataset is the latest revision.

src/app.py
yuanzhou, line 2024: Better to use Flask's `jsonify()` which can serialize python list/dict to JSON and wrap it in a Response with the application/json mimetype.

yuanzhou (changes requested):

src/app.py
```diff
@@ -1953,6 +1955,73 @@ def get_dataset_revision_number(id):
...
+        if j['status'].lower() != DATASET_STATUS_PUBLISHED:
+            user_token = auth_helper_instance.getAuthorizationTokens(request.headers)
+            if not user_in_hubmap_read_group(request) or isinstance(user_token, Response):
+                can_display = False
+                if 'next_revision_uuid' in outputjson[-1]['dataset']:
+                    outputjson[-1]['dataset'].pop('next_revision_uuid')
+        if can_display:
+            outputjson.append(revision)
+        total_revisions = total_revisions + 1
+    outputjson.reverse()
```
yuanzhou, line 2023: We can't rely on this `reverse()` because the list is not guaranteed an ordered list.

yuanzhou (changes requested):

PR #235, opened 2021-09-24
Derek furst/datasets feature revisions

yuanzhou:
Following the merged PR #229 but for TEST

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1955,6 +1955,8 @@ def get_dataset_revision_number(id):
     # Response with the integer
     return jsonify(revision_number)
 
+
+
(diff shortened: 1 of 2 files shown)


yuanzhou:
@DerekFurstPitt has tested on DEV, ready to move to TEST:
>Found a published dataset. Created a new dataset and left "previous revision uuid" as that published dataset. When going to `datasets/<id>/revision` I correctly get back "2". When I go to `/datasets/<id>/revisions` I get the 2 datasets in order of newest first (as is desired). If I send the same request without authorization, I only get back the published dataset and the next_revision_uuid field is missing. I think its working correctly

PR #237, opened 2021-09-27
Added functionality to either include or

DerekFurstPitt:
added the ability to chose to include or not include the full dataset for each revision in the list.


yuanzhou:
Will go to TEST directly since it's a simple change.

yuanzhou (approved):

PR #244, opened 2021-09-30
updated docs

shirey:
- updated /datasets/<id>/retract open-api spec to pass syntax check and added request schema
 - updated /datasets/<id>/revisions open-api spec to pass syntax check and added response schema


DerekFurstPitt (approved):
Looks good. I'm never sure how to format the "schema" portion of a response.

PR #268, opened 2021-12-03
Derek furst/prov info

DerekFurstPitt:
Implemented the datasets/prov-info end point. Will return provenance information about each dataset. By default, returns a tsv file that the user is prompted to download, optionally can use ?format=json to instead return a json of the provenance information.


shirey (changes requested):
@DerekFurstPitt Why did you delete app.cfg.example?  You probably need to put that back....

Also, please update the openapi yaml spec file to include this new method

DerekFurstPitt:
> @DerekFurstPitt Why did you delete app.cfg.example? You probably need to put that back....
> 
> Also, please update the openapi yaml spec file to include this new method

I changed it to app.cfg for local testing, when I went to commit, I must have forgot to uncheck that deletion. I can put it back no problem

DerekFurstPitt:
@shirey I've added the app.cfg.example doc back in. in case you didn't see my comment above, I changed it to app.cfg for local testing and forgot to remove the deletion when creating a commit. I also updated the yaml

yuanzhou (changes requested):
@shirey @DerekFurstPitt because I don't quite understand what the Cypher query does and what this API call does, I just reviewed the PR from the coding perspective. 

@DerekFurstPitt we'll also need to add the new endpoints to Gateway.

PR #269, opened 2021-12-06
added thumbnail support to schema for uploads

jeffechiu:


yuanzhou:
Some questions on this Upload.thumbnail field:
- should we make it `immutable: true` to disallow update via a PUT call?
- is it required field on new Upload creation?
- is this thumbnail field supposed to be reindexed?

yuanzhou:
@jeffechiu I'll close this PR based on our conversation. For the backend changes, you should be able to just copy and past this section to under `Sample` in the yaml: https://github.com/hubmapconsortium/entity-api/blob/test-release/src/schema/provenance_schema.yaml#L360-L389

The rest of the work will be in ingest-ui to facilitate the Sample thumbnail upload/remove.

PR #272, opened 2021-12-10
change revisions method to not specify dataset type

shirey:
Per request from John Conroy:
  - change output to use `uuid` instead of `dataset_uuid`
  - provide a second, not specific dataset route `/entities/<id>/revisions`

### DIFF
--- a/entity-api-spec.yaml
+++ b/entity-api-spec.yaml
@@ -1448,7 +1448,7 @@ paths:
               schema:
                 type: object
                 properties:
-                  dataset_uuid:
+                  uuid:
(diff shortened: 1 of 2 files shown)


yuanzhou (approved):
I also updated the yaml spec to reflect this change: https://github.com/hubmapconsortium/entity-api/commit/538d8c28dee16427fb11b755c5eed079ddef6ceb

I'll just go ahead and add the new endpoint in Gateway as well.

DerekFurstPitt (approved):
Looks good to me

PR #273, opened 2021-12-14
Derek furst/prov info

DerekFurstPitt:


src/app.py
```diff
@@ -2217,46 +2218,61 @@ def get_associated_organs_from_dataset(id):
 """
 @app.route('/datasets/prov-info', methods=['GET'])
 def get_prov_info():
+    # String constants from app.cfg
+    globus_groups_url = app.config['GLOBUS_GROUPS_JSON_URL']
+    organs_list_url = app.config['ORGANS_LIST']
```
yuanzhou, line 2223: Let's create an internal function to load and parse the organ codes instead of using a new config url.

yuanzhou (changes requested):

DerekFurstPitt:
Changes made

PR #279, opened 2021-12-16
Add thumbnail file support to Sample

yuanzhou:
Following merged PR #275 but for TEST


yuanzhou:
The same fix has been merged into test-release via AWS API Gateway changes.

PR #284, opened 2022-01-04
Derek furst/prov info

DerekFurstPitt:
Changed the name of dataset_lab_id to lab_id_or_name. Added code that converts organ codes and data type codes to their
respective descriptions. Added processed dataset uuid, hubmap_id, status, and portal url fields

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2252,6 +2252,10 @@ def get_prov_info():
     HEADER_SAMPLE_METADATA_HUBMAP_ID = 'sample_metadata_hubmap_id'
     HEADER_SAMPLE_METADATA_SUBMISSION_ID = 'sample_metadata_submission_id'
     HEADER_SAMPLE_METADATA_UUID = 'sample_metadata_uuid'
+    HEADER_PROCESSED_DATASET_UUID = 'processed_dataset_uuid'
+    HEADER_PROCESSED_DATASET_HUBMAP_ID = 'processed_dataset_hubmap_id'
+    HEADER_PROCESSED_DATASET_STATUS = 'processed_dataset_status'
+    HEADER_PROCESSED_DATASET_PORTAL_URL = 'processed_dataset_portal_url'
     ASSAY_TYPES_URL = 'https://raw.githubusercontent.com/hubmapconsortium/search-api/master/src/search-schema/data/definitions/enums/assay_types.yaml'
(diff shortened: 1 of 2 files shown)


src/app.py
```diff
@@ -2218,7 +2219,7 @@ def get_associated_organs_from_dataset(id):
 """
 @app.route('/datasets/prov-info', methods=['GET'])
 def get_prov_info():
-    # String constants from app.cfg
+    # String constants
     GLOBUS_GROUPS_URL = 'https://raw.githubusercontent.com/hubmapconsortium/commons/test-release/hubmap_commons/hubmap-globus-groups.json'
```
yuanzhou, line 2223: Let's remove this `GLOBUS_GROUPS_URL` since it's only used in the error message. Considering we have dev-integrate/master branches of commons being used on DEV/PROD, this URL may confuse us, also commons is published to PyPI now.

src/app.py
```diff
@@ -2321,6 +2348,25 @@ def get_prov_info():
         internal_dict[HEADER_DATASET_CREATED_BY_EMAIL] = dataset['created_by_user_email']
         internal_dict[HEADER_DATASET_DATE_TIME_MODIFIED] = datetime.fromtimestamp(int(dataset['last_modified_timestamp']/1000.0))
         internal_dict[HEADER_DATASET_MODIFIED_BY_EMAIL] = dataset['last_modified_user_email']
+        internal_dict[HEADER_DATASET_LAB_ID] = dataset['lab_dataset_id']
```
yuanzhou, line 2351: Do you need to check the existence of `lab_dataset_id` before assigning it? I don't think this field is always available and we'll get KeyError exception when it's not set.

https://github.com/hubmapconsortium/entity-api/blob/Derek-Furst/prov-info/src/schema/provenance_schema.yaml#L250

src/app.py
```diff
@@ -2541,6 +2633,25 @@ def get_prov_info_for_dataset(id):
     internal_dict[HEADER_DATASET_DATE_TIME_MODIFIED] = datetime.fromtimestamp(
         int(dataset['last_modified_timestamp'] / 1000.0))
     internal_dict[HEADER_DATASET_MODIFIED_BY_EMAIL] = dataset['last_modified_user_email']
+    internal_dict[HEADER_DATASET_LAB_ID] = dataset['lab_dataset_id']
```
yuanzhou, line 2636: Same as above comment

yuanzhou (changes requested):

yuanzhou (approved):

PR #293, opened 2022-01-31
Derek furst/prov info fix

DerekFurstPitt:
Fixed 2 bugs. 

1st bug is a 500 error would be raised if ?dataset_status and ?group_uuid were both used. Because of the string manipulation being used, if both were selected, we'd end up with 2 "where" appearing twice in the same statement. This has been fixed. 

2nd bug was that ?dataset_status=QA was returning nothing when it should be returning something. This was because of all the 
status types, QA is the only one that neo4j has as all caps. "Published", "New", "Error" etc were all capitalized only in the first letter
but "QA" is required to be all caps. This has been fixed.


src/app_neo4j_queries.py
```diff
@@ -951,7 +953,15 @@ def get_prov_info(neo4j_driver, param_dict):
...
     if 'dataset_status' in param_dict:
-        dataset_status_query_string = f" WHERE ds.status = '{param_dict['dataset_status'].lower().capitalize()}'"
+        if first_param:
+            print("true")
+            dataset_status_query_string = f" WHERE ds.status = '{param_dict['dataset_status'].lower().capitalize()}'"
+            if param_dict['dataset_status'].lower() == "qa":
+                dataset_status_query_string = f" WHERE ds.status = '{param_dict['dataset_status'].upper()}'"
+        else:
+            dataset_status_query_string = f" AND ds.status = '{param_dict['dataset_status'].lower().capitalize()}'"
+            if param_dict['dataset_status'].lower() == "qa":
```
yuanzhou, line 963: Same as the status check above

yuanzhou (changes requested):

PR #364, opened 2022-06-21
Added function "normalize_status(status)" to schema_manager.py. This …

DerekFurstPitt:
Added function "normalize_status(status)" to schema_manager.py. This method
is called during update_entity to replace the value of status, if present
with the normalized version (lower().capitalize(), except QA which is 
always capitalized).


DerekFurstPitt:
@yuanzhou normalization for status is now being applied to sub_status as well

PR #367, opened 2022-06-27
Derek furst/dbgapurl

DerekFurstPitt:
@yuanzhou This adds dbgap_url to the provenance schema as well as the entity api spec schema. I also took the liberty of adding sub status and retraction reason to the entity api spec because they weren't put there yet. Additionally, I noticed that there was a duplicate description inside an existing property in the api spec so I removed it. 

Looking at the compare/diff screen, it looks as though this merge will add the capitalization changes for some reason so those must not have existed on dev-integrate somehow. I created this branch off of main.


yuanzhou:
Replaced by PR #368

PR #385, opened 2022-07-26
Support multiple search-api instances for reindex

yuanzhou:
- allow multiple search-api urls in config
- use assay type yaml raw file to retrieve description instead of calling search-api to eliminate the complicity of multiple search-api instances
- reindex against each search-api instance
- improved cache mechanism with TTL for performance


yuanzhou:
Link to https://github.com/hubmapconsortium/search-api/pull/558

PR #396, opened 2022-10-05
Derek furst/datasets prov info public

DerekFurstPitt:
Modified the neo4j query for datasets/<id>/prov-info. Now it will work with
non-primary datasets.


yuanzhou:
The changes are simple and can go TEST directly.

yuanzhou (approved):

PR #401, opened 2022-11-15
Include Samples in prov info to support files-api

kburke:
Include sample information with provenance info to support the files-api creating file info documents.
Form complete uuid-api call "in code" from a configured base URL, replacing a configured complete URL.


yuanzhou (changes requested):
@kburke we'll also need to update the uuid-api URL by adding the `/uuid` path when generating new IDs. 

https://github.com/hubmapconsortium/entity-api/blob/kburke/files-api/src/schema/schema_manager.py#L1232

yuanzhou (approved):

PR #404, opened 2022-12-05
Derek furst/identify paired datasets

DerekFurstPitt:
Implemented endpoint `datasets/paired-dataset` which takes a dataset id (hubmap_id or uuid) and a target data type, and will search for paired datasets of that data_type who a first sample ancestor with the given dataset. Optionally, search depth query
parameter sets the max number of generations beneath the sample ancestor to search for the paired dataset. Added accompanying function in app_neo4j_queries.

### DIFF
--- a/entity-api-spec.yaml
+++ b/entity-api-spec.yaml
@@ -2000,3 +2000,45 @@ paths:
           description: THe user is not authorized to use this method
         '500':
           description: Internal error
+  '/datasets/paired_dataset':
+    get:
+      summary: Retrieve uuids for associated dataset of given data_type which shares a sample ancestor of given dataset id
+      parameters:
+        - name: id
+          in: path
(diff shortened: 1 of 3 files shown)


DerekFurstPitt:
Tagging @SFD5311 as an additional reviewer to make sure that this has the desired functionality. 

Basically the endpoint accepts the uuid or hubmap id of the dataset we want to find paired datasets for in the url path. There are 2 query parameters, one optional, one mandatory. The mandatory one is the data_type we're searching for. The optional one is "search depth" which determines how many generations down to search. 

It returns a list with the uuid's of all (if any) matching datasets of the specified data_type that share the nearest sample ancestor with the dataset who's id is provided by the user. 

If this is not the functionality you're looking for or if there's any changes you'd make, let me know and I'll make modifications

SFD5311 (approved):
Looks great.  Thanks!

shirey (approved):
Looks good

shirey:
Approved.

PR #433, opened 2023-01-31
Modified datasets/prov-info and samples/prov-info to check the size o…

DerekFurstPitt:
Modified datasets/prov-info and samples/prov-info to check the size of the 
returned json before returning. If it is at (or near) the max for AWS, 
return a 400 prompting the user to apply some optional parameters to reduce
the size. Changed the date values in datasets/prov-info and 
datasets/<id>/prov-info by converting them to strings from datetime
objects. While they were datetime objects, they weren't valid json, so 
dumping to json would result in a 500.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2857,10 +2857,10 @@ def get_prov_info():
         dataset_prov_list.append(internal_dict)
 
     # Determine whether the size of the returned data exceeds or nearly exceeds the AWS Gateway 10MB maximum size. If it
-    # is greater than 9MB Return a 400 and prompt the user to reduce the size of the output by applying optional
+    # is greater than 9437184 bytes Return a 400 and prompt the user to reduce the size of the output by applying optional
     # argument filters.
     dataset_prov_json_encode = json.dumps(dataset_prov_list).encode('utf-8')
-    if len(dataset_prov_json_encode) > 9000000:
+    if len(dataset_prov_json_encode) > 9437184:
         bad_request_error(
             "Request generated a response over the 10MB limit.  Sub-select the results using a query parameter.")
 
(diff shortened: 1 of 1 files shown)


src/app.py
```diff
@@ -2856,6 +2856,14 @@ def get_prov_info():
         # Each dataset's dictionary is added to the list to be returned
         dataset_prov_list.append(internal_dict)
 
+    # Determine whether the size of the returned data exceeds or nearly exceeds the AWS Gateway 10MB maximum size. If it
+    # is greater than 9MB Return a 400 and prompt the user to reduce the size of the output by applying optional
+    # argument filters.
+    dataset_prov_json_encode = json.dumps(dataset_prov_list).encode('utf-8')
+    if len(dataset_prov_json_encode) > 9000000:
```
yuanzhou, line 2863: Let's keep it consistent by using `9*(2**20)` (9437184) rather than `9000000` https://github.com/hubmapconsortium/uuid-api/blob/main/src/instance/app.cfg.example#L35 which is closer to 10MB. Also update the comment.

src/app.py
```diff
@@ -3448,7 +3455,14 @@ def get_sample_prov_info():
 
         # Each sample's dictionary is added to the list to be returned
         sample_prov_list.append(internal_dict)
-        
+
+    # Determine whether the size of the returned data exceeds or nearly exceeds the AWS Gateway 10MB maximum size. If it
+    # is greater than 9MB Return a 400 and prompt the user to reduce the size of the output by applying optional
+    # argument filters.
+    sample_prov_json_encode = json.dumps(sample_prov_list).encode('utf-8')
+    if len(sample_prov_json_encode) > 9000000:
```
yuanzhou, line 3463: Same as above.

yuanzhou (changes requested):

PR #468, opened 2023-04-10
Entity instanceof

yuanzhou:
Following merged PR #467 but for TEST

New endpoint is already added to API Gateway


yuanzhou:
PR #475 contains all the commits

PR #485, opened 2023-05-24
Support Collections as regular entities

kburke:
Initial revisions to support Collections as regular entities without disrupting existing Collections-only endpoints used for indexing.


src/app.py
```diff
@@ -508,59 +624,40 @@ def get_entity_by_id(id):
...
-                    # Remove the property from the resulting complete_dict
-                    # if the revision is not published
-                    if revision_entity_dict['status'].lower() != DATASET_STATUS_PUBLISHED:
-                        complete_dict.pop(property_to_pop)
-        # Non-published dataset can NOT have next revisions
+        # The user_token is flask.Response on error
+        # Without token, the user can only access public collections, modify the collection result
+        # by only returning public datasets attached to this collection
+        if isinstance(user_token, Response):
+            forbidden_error(f"{normalized_entity_type} for {id} is not accessible with credentials presented.")
```
yuanzhou, line 648: Change "with" to "without"

src/schema/provenance_schema.yaml
```diff
@@ -105,15 +105,31 @@ doi_properties: &doi_properties
     # So we can't define it as `immutable: true`
     type: string
     description: "The doi of the registered entity. e.g. 10.35079/hbm289.pcbm.487"
+    before_property_create_validators:
+     - verify_DOI_pair
+     - halt_update_if_DOI_exists
+     - halt_if_DOI_for_unpublished_dataset
```
yuanzhou, line 111: We should only run this `halt_if_DOI_for_unpublished_dataset` trigger for Collection, and skip it for Dataset.

src/schema/schema_validators.py
```diff
@@ -57,11 +58,127 @@ def validate_application_header_before_entity_create(normalized_entity_type, req
...
+        try:
+            ## The following code duplicates some functionality existing in app.py, in
+            ## query_target_entity(), which also deals with caching. In the future, the
+            ## validation logic shared by this file and app.py should become a utility
+            ## module, shared by validators as well as app.py.  But for now, the code
+            ## is repeated for the following.
+
+            # Get cached ids if exist otherwise retrieve from UUID-API. Expect an
+            # Exception to be raised if not found.
+            dataset_uuid_entity = schema_manager.get_hubmap_ids(id=dataset_uuid)
```
yuanzhou, line 157: When the uuid doesn't exist in UUID MySQL, there's no need to go further and look up in Neo4j.

yuanzhou (changes requested):
@kburke thanks for fixing all the typos. We'll go through the questions/changes.

src/schema/schema_constants.py
```diff
@@ -17,4 +18,14 @@ class SchemaConstants(object):
     ASSAY_TYPES_YAML = 'https://raw.githubusercontent.com/hubmapconsortium/search-api/main/src/search-schema/data/definitions/enums/assay_types.yaml'
 
     # For generating Sample.tissue_type
-    TISSUE_TYPES_YAML = 'https://raw.githubusercontent.com/hubmapconsortium/search-api/main/src/search-schema/data/definitions/enums/tissue_sample_types.yaml'
\ No newline at end of file
+    TISSUE_TYPES_YAML = 'https://raw.githubusercontent.com/hubmapconsortium/search-api/main/src/search-schema/data/definitions/enums/tissue_sample_types.yaml'
+
+    EXPECTED_DOI_PREFIX_WITH_SLASH = 'https://doi.org/'
```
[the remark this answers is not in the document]
kburke: DONE.

src/schema/schema_neo4j_queries.py
```diff
@@ -497,6 +497,55 @@ def link_entity_to_direct_ancestors(neo4j_driver, entity_uuid, direct_ancestor_u
 
         raise TransactionError(msg)
 
+
+"""
+Create or recreate one or more linkages (via Activity nodes) 
+between the target entity node and the direct ancestor nodes in neo4j
```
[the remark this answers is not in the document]
kburke: Done.

src/schema/schema_neo4j_queries.py
```diff
@@ -721,6 +770,81 @@ def get_collection_datasets(neo4j_driver, uuid):
...
+Returns
+-------
+dict
+     A dictionary with an entry for each Dataset in a Collection. The dictionary is
+     keyed by Dataset uuid and contains the Dataset data_access_level.
+"""
+def get_collection_datasets_data_access_levels(neo4j_driver, uuid):
+    results = []
+
+    query = (f"MATCH (d: Dataset)-[:IN_COLLECTION]->(c:Collection) "
```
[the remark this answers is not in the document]
kburke: Done.

src/schema/schema_neo4j_queries.py
```diff
@@ -721,6 +770,81 @@ def get_collection_datasets(neo4j_driver, uuid):
...
+    logger.info("======get_collection_datasets_data_access_levels() query======")
+    logger.info(query)
+
+    with neo4j_driver.session() as session:
+        record = session.read_transaction(execute_readonly_tx, query)
+
+        if record and record[record_field_name]:
+            # Just return the list of values
+            results = record[record_field_name]
+        else:
```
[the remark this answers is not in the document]
kburke: Done.

src/app.py
```diff
@@ -763,26 +856,24 @@ def get_entities_by_type(entity_type):
             if property_key not in result_filtering_accepted_property_keys:
                 bad_request_error(f"Only the following property keys are supported in the query string: {COMMA_SEPARATOR.join(result_filtering_accepted_property_keys)}")
 
-            # Only return a list of the filtered property value of each entity
-            property_list = app_neo4j_queries.get_entities_by_type(neo4j_driver_instance, normalized_entity_type, property_key)
+            if normalized_entity_type == 'Collection':
+                # Only return a list of the filtered property value of each public collection
+                final_result = app_neo4j_queries.get_public_collections(neo4j_driver_instance, property_key)
```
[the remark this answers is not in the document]
kburke: We're leaving this "as is", per a conversation Joe had with Bill today.  New card in the future, if it needs to change.

src/app.py
kburke: This message is specific to no credentials at all being presented (and does not cover bad credentials.)  Change message to `f"{normalized_entity_type} for {id} is not accessible without presenting a token."`

src/schema/schema_validators.py
kburke: I've encountered (and caused!) cases where this occurs on Dev, though I know it shouldn't in Prod.
If double-checking is bad enough to remove a check, I believe the UUID check would go, because I want the Neo4j response to check the entity type.
But I think the extra check provides some sanity during development, and might catch a stray case on other tiers. Do you think it is costly enough or confusing enough it should really come out, @yuanzhou?

yuanzhou: Understood. I'm good with your implementation.

src/schema/provenance_schema.yaml
kburke: Modified `provenance_schema.yaml` and `schema_validators.py` to support validators             `halt_DOI_if_collection_missing_elements` and `halt_if_DOI_for_unpublished_dataset`.

Introduced `registered_doi` and `doi_url` under `ENTITIES: Collection: properties:`, as well as leaving under `doi_properties`, to enable these Collection-specific validations.

yuanzhou (approved):

PR #522, opened 2023-08-18
Add after_update() trigger logic for Collections to handle dataset_uuids

kburke:
Added logic for processing the `dataset_uuids` field of a `PUT` request via trigger logic, fixing defect noted in which Collection Dataset members cannot be updated. Removed

Remove support for `/collections/<collection_uuid>/add-datasets` route.
Remove `app.py` `add_datasets_to_collection()`.
Remove `app_neo4j_queries.py` `add_datasets_to_collection()`.


yuanzhou:
Side note: the `PUT /collections/<collection_uuid>/add-datasets` was never added to AWS API Gateway, so no change is needed there.

yuanzhou (approved):

PR #524, opened 2023-08-24
Derek furst/get siblings

DerekFurstPitt:
Implemented sibling endpoint. Behaves similar to parent and children endpoints. By default, if a sibling is an entity and there is another entity that is a revision of it, it will not be returned. This can be changed with the parameter ?include-old-revisions. Allows filtering on status. Property key = uuid is supported.


src/app.py
```diff
@@ -1829,6 +1829,117 @@ def get_children(id):
...
+    if bool(request.args):
+        for arg_name in request.args.keys():
+            if arg_name not in accepted_args:
+                bad_request_error(f"{arg_name} is an unrecognized argument")
+        property_key = request.args.get('property')
+        status = request.args.get('status')
+        include_revisions = request.args.get('include-old-revisions')
+        if status is not None:
+            status = status.lower()
+            if status not in ['published', 'new', 'qa']:
```
yuanzhou, line 1902: All the defined `Dataset.status` in the schema yaml should be supported here and the in Smart API yaml too:

```
 ['new', 'processing', 'published', 'qa', 'error', 'hold', 'invalid', 'submitted']
 ```

yuanzhou (changes requested):

DerekFurstPitt:
These changes have been applied and pushed. As for the status conditions (qa, new, published, etc) this was copied from elsewhere. I do not recall why at the time we only used these conditions. I'll double check the other place and verify if the status check there is appropriate. 

Should be ready to merge

yuanzhou (approved):

PR #530, opened 2023-09-08
Derek furst/track status

DerekFurstPitt:
Implemented property status_history. Its a list containing dictionaries which are status change events.


src/schema/schema_triggers.py
```diff
@@ -556,6 +556,43 @@ def update_file_descriptions(property_key, normalized_type, user_token, existing
...
+            status_history_string = existing_data_dict['status_history'].replace("'", "\"")
+            new_status_history += json.loads(status_history_string)
+    current_time = int(datetime.now().timestamp() * 1000)
+    if 'email' not in new_data_dict:
+        raise KeyError("Missing 'email' key in 'new_data_dict' during calling 'set_status_hisotry()' trigger method.")
+    email = new_data_dict['email']
+    status_entry['status'] = status
+    status_entry['changed_by_email'] = email
+    status_entry['change_timestamp'] = current_time
+    new_status_history.append(status_entry)
```
yuanzhou, line 590: The status history list is not guaranteed to be sorted by the event change time, so we'll need to add `new_status` and `previous_status` as part of each status change event.

yuanzhou (changes requested):
@DerekFurstPitt

src/schema/schema_triggers.py
yuanzhou: Resolved between @DerekFurstPitt and @shirey 
>So rather than doing the limited implementation now and doing a full list down the road, it just made sense to do it like a list now.

src/app.py
```diff
@@ -1363,12 +1364,14 @@ def update_entity(id):
...
             associated_collection_dict = query_target_entity(json_data_dict['associated_collection_uuid'], user_token)
+        if ('status' in json_data_dict) and (json_data_dict['status']):
+            has_updated_status = True
 
         # Generate 'before_update_trigger' data and update the entity details in Neo4j
         merged_updated_dict = update_entity_details(request, normalized_entity_type, user_token, json_data_dict, entity_dict)
 
         # Handle linkages update via `after_update_trigger` methods
-        if has_direct_ancestor_uuids or has_associated_collection_uuid:
+        if has_direct_ancestor_uuids or has_associated_collection_uuid or has_updated_status:
```
yuanzhou, line 1374: This `has_updated_status` flag and check should also be added inside the `elif normalized_entity_type == 'Upload':` section, otherwise the defined `Upload.status.after_update_trigger.set_status_history` will not execute.

yuanzhou (changes requested):

DerekFurstPitt:
Updated to accommodate uploads. In addition to adding `or_has_updated_status` under uploads, I would also need to include the logic that sets that property as true, so instead I just moved that block up one level so it will work for datasets/publications and uploads. Performed some tests locally, uploads work and datasets/publications appear to still be working.

yuanzhou (approved):

PR #538, opened 2023-09-26
Derek furst/add creation action

DerekFurstPitt:
Added property creation_action to datasets and publications. This property is transient and immutable, so it won't be saved in the
database within the dataset/publication node itself, however the trigger method for the activity node with this property has been updated to grab this value if its in the json. If this field isn't present, it defaults to the existing value. Validator verifies that the property is only 'Central Process', 'Lab Process', or 'Multi-assay Split'.

### DIFF
--- a/src/schema/schema_triggers.py
+++ b/src/schema/schema_triggers.py
@@ -1907,7 +1907,8 @@ str: The creation_action string
 def set_activity_creation_action(property_key, normalized_type, user_token, existing_data_dict, new_data_dict):
     if 'normalized_entity_type' not in new_data_dict:
         raise KeyError("Missing 'normalized_entity_type' key in 'existing_data_dict' during calling 'set_activity_creation_action()' trigger method.")
-    
+    if new_data_dict and new_data_dict['creation_action']:
+        return property_key, new_data_dict['creation_action']
     return property_key, f"Create {new_data_dict['normalized_entity_type']} Activity"
 
 
--- a/src/schema/schema_triggers.py
+++ b/src/schema/schema_triggers.py
@@ -1907,8 +1907,8 @@ str: The creation_action string
 def set_activity_creation_action(property_key, normalized_type, user_token, existing_data_dict, new_data_dict):
     if 'normalized_entity_type' not in new_data_dict:
         raise KeyError("Missing 'normalized_entity_type' key in 'existing_data_dict' during calling 'set_activity_creation_action()' trigger method.")
-    if new_data_dict and new_data_dict['creation_action']:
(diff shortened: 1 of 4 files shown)


src/schema/schema_triggers.py
```diff
@@ -1907,7 +1907,8 @@ def get_upload_datasets(property_key, normalized_type, user_token, existing_data
 def set_activity_creation_action(property_key, normalized_type, user_token, existing_data_dict, new_data_dict):
     if 'normalized_entity_type' not in new_data_dict:
         raise KeyError("Missing 'normalized_entity_type' key in 'existing_data_dict' during calling 'set_activity_creation_action()' trigger method.")
-    
+    if new_data_dict and new_data_dict['creation_action']:
```
yuanzhou, line 1910: This condition will cause `KeyError: 'creation_action'` for when `creation_action` field is not provided in the request. Need to check the existence of the key.

src/schema/schema_validators.py
```diff
@@ -509,6 +509,27 @@ def validate_publication_date(property_key, normalized_entity_type, request, exi
...
+request: Flask request object
+    The instance of Flask request passed in from application request
+existing_data_dict : dict
+    A dictionary that contains all existing entity properties
+new_data_dict : dict
+    The json data in request body, already after the regular validations
+"""
+def validate_creation_action(property_key, normalized_entity_type, request, existing_data_dict, new_data_dict):
+    creation_action = new_data_dict[property_key]
+    if creation_action and creation_action not in ["Central Process", "Lab Process", "Multi-Assay Split"]:
```
yuanzhou, line 530: I would make this string match case-insensitive so it's more fault-tolerant, just like the way we check `status` values. It's also a good idea to tell the API consumers what are the accepted values in the `ValueError()` next line. Because these values are not as easy-to-get-right-for-the-first-time as those simple status values.

yuanzhou (changes requested):

src/schema/schema_validators.py
DerekFurstPitt: @yuanzhou I changed `new_data_dict['creatoin_action']` to `new_data_dict.get('creation_action') which on its own should fix the error being raised. While I think it might be acceptable to just use the default creation action when the field is empty, I still went ahead and added validation to make sure its not empty anyway. I made the validation case insensitive and added some case normalization within the trigger so it will always display with the first letters capitalized (e.g. "Lab Process"). I modified the value error message to include the accepted values for creation_action as well.

yuanzhou (approved):

PR #542, opened 2023-10-05
Enable re-index of Collections

kburke:
Enable re-index of Collections during POST and PUT calls, in this service and also search-api/search-adaptor.

Remove markup for old `/collections/<id>/add-datasets` endpoint.

Add a new `/visibility/<id>` endpoint not exposed by AWS Gateway, for use by search-api.  Allows this service to encapsulate criteria for Collection and other entities to be available to the public.

Fix some service startup code to be a little more forthcoming about configuration problems on startup.


src/app.py
```diff
@@ -1464,17 +1524,11 @@ def update_entity(id):
...
-        logger.log(logging.DEBUG
-                   ,f"Skipping re-indexing {entity_dict['entity_type']}"
-                    f" with UUID {entity_dict['uuid']}"
-                    f" until supported by search-api.")
-    else:
-        logger.log(logging.INFO
-                   ,f"Re-indexing for creation of {entity_dict['entity_type']}"
-                    f" with UUID {entity_dict['uuid']}")
+    logger.log(logging.INFO
+               ,f"Re-indexing for creation of {entity_dict['entity_type']}"
```
yuanzhou, line 1528: @kburke the logging message needs a bit tweak here since the entity gets updated in this `update_entity()`

yuanzhou (changes requested):
@kburke thanks for improving the code, just a tiny logging message correction is requested.

yuanzhou (approved):

PR #545, opened 2023-10-09
Derek furst/multiple components datasets

DerekFurstPitt:
Implemented datasets/components endpoint to register multiple components from a mult-assay ancestor. These entities are not
created as part of a typical entity creation process. We can't use the normal triggers because those are configured to work with a single ancestor per child dataset. Instead, all of the validation is handled within this endpoint. For activity creation and linkages to the direct ancestor, new functions were created. While, for now, we are only allowing a single parent ancestor, the functions were created such that they would accommodate multiple ancestors. The main difference between this schema function and the existing ones is that it accepts multiple child entities rather than just one. The activity created is of the type Multi-Assay Split


yuanzhou (changes requested):
@DerekFurstPitt the way you handle this specialized endpoint with mixed use of schema components can be simplified and the SenNet project will appreciate this effort too. I'll describe the steps:

1. First check the required `X-Hubmap-Application` header. And group all the input json validations at the beginning without using any `schema_manager.validate_json_data_against_schema()`. The purpose is to ensure data integrity by checking the required fields and their data values in terms of existence, size, and type...

2. Similar to `create_multiple_samples_details()`, we create two set of IDs (which also validates the `group_uuid` comprehensively) via uuid-api using `schema_manager.create_hubmap_ids()` for the two component datasets. Then build two merged dicts and call `schema_manager.generate_triggered_data()` on each to generate all the dataset node properties via the schema. 

3. As what you've already done, generate one Activity node to be linked to the direct ancestor dataset and the two new component datasets. It's also the time to update the `Activity.creation_action` with the value of "Multi-Assay Split" as you did.

4. Create a similar app neo4j query as `app_neo4j_queries.create_multiple_samples()` and get everything (nodes and relationships) into Neo4j, without using the schema.

5. Add the new datasets to indices via reindex. And return the desired data structure back to ingest-api. Just like what you did.

By following the above approach, we'll have one specialized endpoint that handles the custom json input with custom validations, and one app-specific neo4j query.

src/app.py
```diff
@@ -3785,6 +3786,172 @@ def paired_dataset(id):
...
+    # than a 500 error later if the token is not good.
+    validate_token_if_auth_header_exists(request)
+    # Get user token from Authorization header
+    user_token = get_user_token(request)
+    try:
+        schema_validators.validate_application_header_before_entity_create("Dataset", request)
+    except Exception as e:
+        bad_request_error(str(e))
+
+    require_json(request)
```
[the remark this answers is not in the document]
yuanzhou: No change needed here. False positive.

maxsibilla (approved):
I think this addresses the major points that we all discussed. I believe you also addressed all the issues @yuanzhou  mentioned in the card but I will leave that to him to confirm

src/app.py
```diff
@@ -3785,6 +3786,149 @@ def paired_dataset(id):
...
+         - type: str
+         - description: the uuid for the parent multi assay dataset
+        datasets
+         - type: dict
+         - description: the datasets to be created. Only difference between these and normal datasets are the field "dataset_link_abs_dir"
+
+Returns
+--------
+json array
+    List of uuids of the newly created component datasets
```
yuanzhou, line 3812: We are returning two dicts in a list, not just uuids

src/app.py
```diff
@@ -3785,6 +3786,149 @@ def paired_dataset(id):
...
+
+        # Add back in dataset_link_abs_dir
+        dataset['dataset_link_abs_dir'] = dataset_link_abs_dir
+
+    dataset_list = create_multiple_component_details(request, "Dataset", user_token, json_data_dict.get('datasets'), json_data_dict.get('creation_action'))
+
+    # We wait until after the new datasets are linked to their ancestor before performing the remaining post-creation
+    # linkeages. This way, in the event of unforseen errors, we don't have orphaned nodes.
+    for dataset in dataset_list:
+        schema_triggers.set_status_history('status', 'Dataset', user_token, dataset, {})
```
yuanzhou, line 3890: Good handling of updating the `status_history` by calling the "after_create_trigger" explicitly here!

src/app.py
```diff
@@ -4303,6 +4447,139 @@ def create_multiple_samples_details(request, normalized_entity_type, user_token,
...
+    The user's globus groups token
+json_data_dict: dict
+    The json request dict from user input
+creation_action : str
+    The creation action for the new activity node.
+
+Returns
+-------
+list
+    A list of all the newly generated ids via uuid-api
```
yuanzhou, line 4469: We are not just returning the ids, please update.

yuanzhou (changes requested):
@DerekFurstPitt the rework looks great! In addition to updating the comments to indicate the correct return results, can you also sync the latest `main` into your branch? When I tested locally with your branch, I noticed the dependency issue that had been addressed in `main`.

yuanzhou (approved):

yuanzhou (approved):

PR #550, opened 2023-10-16
Fixed a bug in a group uuid exception handler where a reference is made

DerekFurstPitt:
Fixed a bug in a group uuid exception handler where a reference is made to json_data_dict rather than the first json_data_dict in json_data_dict_list (json_data_dict[0] and json_data_dict[i] where appropriate).

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -4479,25 +4479,6 @@ def create_multiple_component_details(request, normalized_entity_type, user_toke
         # we only need the json data from one of the datasets. The info will be the same for both, so we just grab the first in the list
         new_ids_dict_list = schema_manager.create_hubmap_ids(normalized_entity_type, json_data_dict_list[0], user_token, user_info_dict, len(json_data_dict_list))
     # When group_uuid is provided by user, it can be invalid
-    except schema_errors.NoDataProviderGroupException:
-        # Log the full stack trace, prepend a line with our message
-        if 'group_uuid' in json_data_dict_list[0]:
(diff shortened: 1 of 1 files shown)


yuanzhou (changes requested):
@DerekFurstPitt good catch. I just realized that we don't need to handle the groups exceptions at all during IDs creation. Because those exceptions would never apply to dataset, they only apply to Donor and Upload: https://github.com/hubmapconsortium/entity-api/blob/Derek-Furst/multiple-components-datasets/src/schema/schema_manager.py#L1299-L1343, that's also why I didn't see any error when I tested locally. 

The group checks are taken care of during calling the "before_create_trigger" and that's sufficient. Those error handlings in `create_multiple_samples_details()` would never kick in neither during IDs creation for samples, but we don't want to touch the code for now.

Let's remove the data provider exceptions handling here for multi-components, which makes the code shorter and easy to understand.

DerekFurstPitt:
@yuanzhou Removed those additional exception types. Since those checks were used in the multiple samples endpoints, it didn't seem to hurt including them, but you're right that if those situations can never occur here, its better to leave them out. I also after talking with @maxsibilla about the interaction between ingest api and entity api, I've decided to add a check to datasets in entity api to verify that `dataset_link_abs_dir` is included. Other than making sure it exists, I'm not doing any further checks here, I'll add those checks and error handling on the ingest-api side. I've tested these latest changes locally and it doesn't seem to change much so I think if I can do one more pass on dev I'll be good to move this further along

yuanzhou (approved):

PR #554, opened 2023-10-18
Reapplying fix (json_data_dict -> json_data_dict_list[i]) in exceptio…

DerekFurstPitt:
Reapplying fix (json_data_dict -> json_data_dict_list[i]) in exception. Fix was unintentionally removed when re-adding exceptions

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -4514,7 +4514,7 @@ def create_multiple_component_details(request, normalized_entity_type, user_toke
             internal_server_error(msg)
         except schema_errors.NoDataProviderGroupException:
             # Log the full stack trace, prepend a line with our message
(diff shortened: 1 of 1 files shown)


DerekFurstPitt:
Initially, this error was identified and when I fixed it, I fixed it in 2 places it occurred, once at the top of the function, and once lower down just after `generated_before_create_trigger_data_dict`. It was brought to my attention that some of these exceptions were unnecessary because they only applied to other entity types. So I removed these exceptions. This was premature as it turned out only the first block needed removed, the second block could come in handy. But when I re-added the second exception block, I mistakenly omitted the original fix. This re-applies that fix.

yuanzhou (approved):

Issue #534, opened 2023-09-19
Use `Activity.creation_action` to find derived/processed datasets


shirey:
Anywhere we're querying Neo4j (might be outside of entity-api too..?) Use the new values stored in `Activity.creation_action` to query for derived/processed datasets using the new values `Lab Process` and `Central Process`.  For example to find all derived/processed data you can do something like:
```
match (a:Activity)-[:ACTIVITY_OUTPUT]->(ds:Dataset) where a.creation_action ends with 'Process' return ds.uuid
```

PR #559, opened 2023-10-25
Derek furst/update derived queries

DerekFurstPitt:
Changed dataset_is_primary to use creation_action to determine whether a dataset is primary rather than inspecting its distance from the nearest sample ancestor


yuanzhou:
@DerekFurstPitt The query updates handle the "Central Process" and "Lab Process" Activity types. Do we need to exclude the "Multi-Assay Split" type as well for those component datasets?

DerekFurstPitt:
> @DerekFurstPitt The query updates handle the "Central Process" and "Lab Process" Activity types. Do we need to exclude the "Multi-Assay Split" type as well for those component datasets?

I am uncertain. Issue [534](https://github.com/hubmapconsortium/entity-api/issues/534) only says lab process and central process, and the example query bill included specifically uses the "ends with process" bit

yuanzhou:
From Bill:
>The result of a “Multi-Assay Split” dataset isn’t a derived/processed dataset so should not be included in this task.

src/app_neo4j_queries.py
```diff
@@ -626,8 +579,9 @@ def get_prov_info(neo4j_driver, param_dict, published_only):
     if published_only:
         published_only_query_string = f" AND toUpper(ds.status) = 'PUBLISHED'"
         published_only_revisions_string = f" WHERE toUpper(rev.status) = 'PUBLISHED'"
-    query = (f"MATCH (ds:Dataset)<-[:ACTIVITY_OUTPUT]-(a)<-[:ACTIVITY_INPUT]-(firstSample:Sample)<-[*]-(donor:Donor)"
-             f"WHERE not (ds)-[:REVISION_OF]->(:Dataset)"
+    query = (f"MATCH (ds:Dataset)<-[:ACTIVITY_OUTPUT]-(a:Activity)<-[*]-(firstSample:Sample)<-[*]-(donor:Donor)"
+             f" WHERE not (ds)-[:REVISION_OF]->(:Dataset)"
+             f" AND NOT toLower(a.creation_action) ENDS WITH 'process'"
```
yuanzhou, line 584: The cypher query Bill used in https://github.com/hubmapconsortium/entity-api/issues/534 only finds the derived/processed datasets. However, you are using negation here, so you may need to double check and exclude the “Multi-Assay Split”.

yuanzhou (changes requested):

src/app_neo4j_queries.py
yuanzhou: Confirmed with Bill:
>“Multi-Assay Split” datasets are primary datasets.

yuanzhou (approved):

PR #565, opened 2023-11-13
Derek furst/multiple revisions

DerekFurstPitt:
Updated the property previous_revision_uuid (singular) to allow for either a list of previous_revisions or a single previous revision. The validation has changed. It used to not be allowed to add revisions to a dataset/publication that already had revisions of it. Now it is possible, however only if the existing revisions are the most recent ones. If a revision has been made of THAT revision, adding new revisions off of the earlier generation is disallowed. Also, an additional validation has been added when a list of multiple previous_revisions is given: revisions in the list may not be revisions of each other.


src/app.py
```diff
@@ -983,26 +983,32 @@ def create_entity(entity_type):
...
+                nested_revisions =  app_neo4j_queries.nested_previous_revisions(neo4j_driver_instance, previous_revision_list)
+                if nested_revisions:
+                    bad_request_error(f"{nested_revisions[0][0]} is a revision of {nested_revisions[1][0]}. Datasets in previous_revision_uuid must not be revisions of eachother")
+            else:
+                previous_revision_list = [json_data_dict['previous_revision_uuid']]
+            for previous_revision in previous_revision_list:
+                previous_version_dict = query_target_entity(previous_revision, user_token)
+
+                # Make sure the previous version entity is either a Dataset or Sample (and publication 2/17/23)
+                if previous_version_dict['entity_type'] not in ['Sample'] and \
```
yuanzhou, line 998: Not introduced by your changes. But let's disallow `Sample` type, only allow `Dataset` (`Publication` too since it's a subclass of `Dataset`). We've never had Sample as the previous revision. And with the multi-revisions feature, it only applies to Dataset.

src/app_neo4j_queries.py
```diff
@@ -399,6 +459,76 @@ def get_next_revisions(neo4j_driver, uuid, property_key = None):
...
+                 # apoc.coll.toSet() reruns a set containing unique nodes
+                 f"RETURN apoc.coll.toSet(COLLECT(next.{property_key})) AS {record_field_name}")
+    else:
+        query = (f"MATCH (e:Entity)<-[:REVISION_OF*]-(parent:Entity)<-[:REVISION_OF*]-(next:Entity) "
+                 f"WHERE e.uuid='{uuid}' "
+                 # COLLECT() returns a list
+                 # apoc.coll.toSet() reruns a set containing unique nodes
+                 f"RETURN apoc.coll.toSet(COLLECT(next)) AS {record_field_name}")
+
+    logger.info("======get_next_revisions() query======")
```
yuanzhou, line 479: Update the log with using the `is_next_revision_latest()`

src/app_neo4j_queries.py
```diff
@@ -399,6 +459,76 @@ def get_next_revisions(neo4j_driver, uuid, property_key = None):
...
+    Else return None
+"""
+def nested_previous_revisions(neo4j_driver, previous_revision_list):
+    query = (f"WITH {previous_revision_list} AS uuidList "
+            "MATCH (ds1:Dataset)-[r:REVISION_OF]->(ds2:Dataset) "
+            "WHERE ds1.uuid IN uuidList AND ds2.uuid IN uuidList "
+            "WITH COLLECT(DISTINCT ds1.uuid) AS connectedUUID1, COLLECT(DISTINCT ds2.uuid) as connectedUUID2 "
+            "RETURN connectedUUID1, connectedUUID2 ")
+
+    logger.info("======nexted_previous_revisions() query======")
```
yuanzhou, line 521: Typo, change `nexted` to `nested`

src/schema/provenance_schema.yaml
```diff
@@ -645,7 +645,9 @@ ENTITIES:
         description: "The displayname of globus group which the user who created this entity is a member of"
         before_create_trigger: set_group_name #same as group_uuid, except set group_name
       previous_revision_uuid:
-        type: string
+        type:
+          - string
+          - list
```
yuanzhou, line 650: You only added this multi data types to `Publication`, will need it for `Dataset.previous_revision_uuid` too.

src/schema/schema_neo4j_queries.py
```diff
@@ -652,15 +652,16 @@ def link_collection_to_datasets(neo4j_driver, collection_uuid, dataset_uuid_list
...
     The uuid of previous revision entity
 """
-def link_entity_to_previous_revision(neo4j_driver, entity_uuid, previous_revision_entity_uuid):
+def link_entity_to_previous_revision(neo4j_driver, entity_uuid, previous_revision_entity_uuids):
     try:
-        with neo4j_driver.session() as session:
-            tx = session.begin_transaction()
+        for previous_uuid in previous_revision_entity_uuids:
+            with neo4j_driver.session() as session:
+                tx = session.begin_transaction()
```
yuanzhou, line 660: We should use the for loop inside the transaction of a session, rather than open a new session and new transaction for each iteration.

src/app_neo4j_queries.py
```diff
@@ -399,6 +459,76 @@ def get_next_revisions(neo4j_driver, uuid, property_key = None):
 
     return results
 
+
```
yuanzhou, line 462: Missing comment block for this `is_next_revision_latest()` method

src/app_neo4j_queries.py
```diff
@@ -399,6 +459,76 @@ def get_next_revisions(neo4j_driver, uuid, property_key = None):
...
+def is_next_revision_latest(neo4j_driver, uuid, property_key = None):
```
yuanzhou, line 463: We don't need this `property_key` as an optional argument in the method signature at all. Modify the query to only return a list of uuids, then move the non-empty list check here, and return a boolean as the final result. When function is named as `is_xxx` a boolean return is expected as the norm.

yuanzhou (changes requested):
@DerekFurstPitt, great handling to support both string and list for `previous_revision_uuid` field! Let's address a few minor tweaks then we'll get it on DEV.

yuanzhou (approved):

Issue #551, opened 2023-10-17
Expose `creation_action` attribute from activity on Dataset GET


shirey:
On a `GET /entities/<dataset id>` call, when getting the information for a Dataset entity include the `creation_action` in the response. Unlike all other Dataset attributes `creation_action` is stored in the associated action that represents the creation of that dataset (associated via the ACTIVITY_OUTPUT relationship).

An example Neo4j query that will return the `creation_action` for a Dataset, given the Dataset UUID is:
```
match (ds {uuid:'<dataset uuid>'})<-[:ACTIVITY_OUTPUT]-(a:Activity) return a.creation_action;
```

ChuckKollar:
Using entity_dict['creation_action_activity'] as Entities.Dataset already has a creation_action property which is processed in a different manner see provenance_schema.yaml and code in this file.

ChuckKollar:
Also for Publication.

PR #574, opened 2023-11-28
Add Dataset Activity.creation_action to ENTITY.Dataset as creation_action_activity

ChuckKollar:
Issue https://github.com/hubmapconsortium/entity-api/issues/551

### DIFF
--- a/src/instance/app.cfg.example
+++ b/src/instance/app.cfg.example
@@ -28,6 +28,10 @@ UUID_API_URL = 'http://uuid-api:8080'
 # Works regardless of the trailing slash
 INGEST_API_URL = 'https://ingest-api.dev.hubmapconsortium.org'
 
+# URL for talking to Ontology API (default for DEV)
+# Works regardless of the trailing slash
+ONTOLOGY_API_URL = 'https://ontology-api.dev.hubmapconsortium.org'
+
 # A list of URLs for talking to multiple Search API instances (default value used for docker deployment, no token needed)
 # Works regardless of the trailing slash /
 SEARCH_API_URL_LIST = ['http://search-api:8080']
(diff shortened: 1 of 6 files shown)


shirey (changes requested):
@yuanzhou instead of adding adding a new attribute would an `on_read_trigger` on the existing `creation_action` attribute which fills in the value work?

@ChuckKollar If possible I'd like to use the same `creation_action` attribute that exists already, but only currently used for POST operations, I think..

DerekFurstPitt:
@shirey That was my thinking as well reading through this. If we still want this property to remain transient (for datasets) but want it to be returned on a GET request, I think an on-read trigger would suffice and wouldn't necessitate creating another transient property.

DerekFurstPitt (approved):
With these changes, this looks good to me.

ChuckKollar:
OK! Great! Thanks for the suggestions and the read over! :-)

shirey (approved):

src/instance/app.cfg.example
```diff
@@ -28,6 +28,10 @@ UUID_API_URL = 'http://uuid-api:8080'
 # Works regardless of the trailing slash
 INGEST_API_URL = 'https://ingest-api.dev.hubmapconsortium.org'
 
+# URL for talking to Ontology API (default for DEV)
+# Works regardless of the trailing slash
+ONTOLOGY_API_URL = 'https://ontology-api.dev.hubmapconsortium.org'
```
yuanzhou, line 33: @ChuckKollar is this change relying on the UBKG ontology-api? I had a branch that introduced the UBKG stuff, not sure how this config got into your code.

yuanzhou (changes requested):

ChuckKollar:
This PR did contain the addition of the following code in src/instance/app.cfg.example
# URL for talking to Ontology API (default for DEV)
# Works regardless of the trailing slash
ONTOLOGY_API_URL = 'https://ontology-api.dev.hubmapconsortium.org'

But Joe wanted it removed, so I did.

yuanzhou (approved):

PR #583, opened 2023-12-05
Kburke/add dataset type attribute

kburke:


src/schema/schema_validators.py
```diff
@@ -41,6 +43,36 @@ def validate_application_header_before_entity_create(normalized_entity_type, req
...
+    The instance of Flask request passed in from application request
+existing_data_dict : dict
+    A dictionary that contains all existing entity properties
+new_data_dict : dict
+    The json data in request body, already after the regular validations
+"""
+def validate_recognized_dataset_type(property_key, normalized_entity_type, request, existing_data_dict, new_data_dict):
+    # If the proposed Dataset dataset_type ends with something in square brackets, anything inside
+    # those square brackets are acceptable at the end of the string.  Simply validate the start.
+    proposed_dataset_type_prefix = re.sub(pattern='[ ]*\[.*]$', repl='', string=new_data_dict['dataset_type'])
```
yuanzhou, line 67: @kburke there should be only one space allowed before the square brackets. Can you update the regex pattern? This would help climate malformed dataset types down the road. And the current thinking is that the Rules Engine (PSC work) will provide a formatted description for a processed dataset, so if they don't format the type correctly, we'll be able to find out and tell them to correct.

yuanzhou (changes requested):

src/schema/schema_validators.py
kburke: @yuanzhou I considered that, but not being aware of any such requirement, I was accepting zero or more.
This made it much easier because I didn't have to use a substitution group and could focus on the area adjacent to the square brackets.
But if the requirement is exactly one, I will work toward `pattern='\([^ ]\) \[.*]$', repl='\1'`...

yuanzhou: @kburke Yes, only one space. There are also dataset types like `scRNA-seq (10x Genomics) [Salmon]` and we don't want to go too deep to validate the `()` section. Only care about the `[]` for now. Hopefully eventually all the dataset types will be defined in UBKG and we won't need to validate here.

kburke: Using the pattern `re.sub(pattern='(\S)\s\[.*\]$', repl=r'\1', ... `, below are some samples of the error message posted, when the string inside the single quotes is specified.  These patterns were accepted prior to @yuanzhou 's request.

```
{
    "error": "400 Bad Request: Proposed Dataset dataset_type 'RNASeq[salmon]' is not recognized in the existing ontology. Valid values are: ['Histology', 'Molecular Cartography', 'RNASeq', 'ATACSeq', 'SNARE-seq2', 'PhenoCycler', 'CyCIF', 'MERFISH', 'MALDI', '2D Imaging Mass Cytometry', 'nanoSPLITS', 'Auto-fluorescence', 'Confocal', 'Thick section Multiphoton MxIF', 'Second Harmonic Generation (SHG)', 'Enhanced Stimulated Raman Spectroscopy (SRS)', 'SIMS', 'Cell DIVE', 'CODEX', 'Lightsheet', 'MIBI', 'LC-MS', 'DESI', '10x Multiome', 'Visium']."
}
```

```
{
    "error": "400 Bad Request: Proposed Dataset dataset_type 'RNASeq     [salmon]' is not recognized...
```

yuanzhou (approved):

Issue #589, opened 2023-12-14
Convert `dataset_type` to a required field and no longer require `data_types`


shirey:
For both Publication and Dataset, require `dataset_type` to be set on creations and no longer require `data_types` to be set.

Do this work and test, but hold it in a branch named: `soft-assay-conversion` as it will need to be released in coordination with other changes across other services/applications.

PR #594, opened 2024-01-10
Derek furst/make dataset type required

DerekFurstPitt:
Changed dataset types to be required on create (but not update) and changed data_types to be not required. These changes are for dataset and publication


yuanzhou (changes requested):
@DerekFurstPitt it was not described explicitly in https://github.com/hubmapconsortium/entity-api/issues/589. Once we make `data_types` optional and `dataset_type` a required field, the auto generated field `Dataset.dataset_title` via a read trigger should also be using `dataset_type`. And there's no need to [convert based on the old field](https://github.com/hubmapconsortium/entity-api/blob/dd14d49d30cb8ba5623dd983fc679b074d2e5fba/src/schema/schema_triggers.py#L1008-L1018).

PR #596, opened 2024-01-12
modified prov-info and <id>/prov-info to use dataset_type instead of …

DerekFurstPitt:
modified prov-info and <id>/prov-info to use dataset_type instead of data
types. samples/prov-info didn't appear to include any references. Also 
applied an identical change to what was done in the data-sankey branch to 
make validate_organ_code accept an optional argument that includes the organ
dictionary so that ontology api doesn't need called thousands of times.


src/app.py
```diff
@@ -2721,7 +2719,7 @@ def get_prov_info():
     # Token is not required, but if an invalid token is provided,
     # we need to tell the client with a 401 error
     validate_token_if_auth_header_exists(request)
-
+    organ_types_dict = schema_manager.get_organ_types()
```
yuanzhou, line 2722: Looks like the `organ_types_dict` is assigned twice with the same call: https://github.com/hubmapconsortium/entity-api/blob/95b243e3da97d3c9138507ca598e64acb056cf28/src/app.py#L2722-L2728

Just keep one.

src/app.py
```diff
@@ -3016,7 +2992,7 @@ def get_prov_info_for_dataset(id):
     # Token is not required, but if an invalid token provided,
     # we need to tell the client with a 401 error
     validate_token_if_auth_header_exists(request)
-
+    organ_types_dict = schema_manager.get_organ_types()
```
yuanzhou, line 2995: Found duplicate in `/datasets/<id>/prov-info` too

yuanzhou (changes requested):

src/app.py
yuanzhou: I don't see the duplicates in the sankey data PR, could just be in this prov-info branch. 

That's good performance optimization you did even though that organ call internally is being cached.

DerekFurstPitt: Yes my concern was less about the speed of the prov-info endpoints (since as you mentioned, its cached) but rather clogging up ontology with thousands of requests.

DerekFurstPitt:
@yuanzhou those duplicates have been removed

PR #606, opened 2024-01-30
Derek furst/tuplets

DerekFurstPitt:
Implemented tuplets endpoint. very similar to siblings endpoint, but rather than sharing a direct ancestor, they share the same
exact parent activity node. Some small tweaks were required. Parameter "include revisions" removed. As these tuplets are spawned from the same process, they will never be revisions of each other.


src/app.py
```diff
@@ -1825,6 +1825,109 @@ def get_siblings(id):
...
+    if property_key is not None:
+        return jsonify(tuplet_list)
+    # Generate trigger data
+    # Skip some of the properties that are time-consuming to generate via triggers
+    # Also skip next_revision_uuid and previous_revision_uuid for Dataset to avoid additional
+    # checks when the target Dataset is public but the revisions are not public
+    properties_to_skip = [
+        # Properties to skip for Sample
+        'direct_ancestor',
+        # Properties to skip for Dataset
```
yuanzhou, line 1913: Recently we added `creation_action` to Dataset and Publication, which is generated/returned via a `on_read_trigger`. This would potentially cause performance issue due to extra neo4j loopup when there are hundreds of tuplets. We should skip this property.

Similarly, also skip the following: `local_directory_rel_path`, `previous_revision_uuids`, and `next_revision_uuids` (from recent multi-revisions work)

yuanzhou (changes requested):
@DerekFurstPitt when you make these tweaks, let's apply the same to the existing `/siblings/<id>` call by:

- Rename it to `/entities/<id>/siblings`
- Skip those additional triggers

src/app.py
DerekFurstPitt: @yuanzhou Requested changes applied

PR #609, opened 2024-02-02
Derek furst/update siblings

DerekFurstPitt:
changed siblings path from /siblings/<id> to /entities/<id>/siblings


DerekFurstPitt:
Added new properties to skip

yuanzhou (approved):

Issue #604, opened 2024-01-25
Dataset Id sample and donor aggregate endpoints


ChuckKollar:
Create one new endpoints in `entity-api` that will return the `samples`, `organs`, and `donors` uuids of dataset id given.

These endpoints will use the Neo4J queries documented in https://github.com/hubmapconsortium/ingest-api/issues/411

This endpoint will be used in: https://github.com/hubmapconsortium/ingest-api/issues/411

Remember to update the OpenAPI .yaml file.

ChuckKollar:
See PR: https://github.com/hubmapconsortium/entity-api/pull/612

yuanzhou:
Update: 
Instead of creating a new endpoint that returns all the uuids, we create two new endpoints `/datasets/<id>/donors` and `/datasets/<id>/samples`. The implementation is similar to the existing `/datasets/<id>/organs` but with modified neo4j queries so we return all the entity dicts rather than just uuids. 

For future when needed, we can add uuid filtering to only return uuids to the new endpoints and the existing `/datasets/<id>/organs`

PR #612, opened 2024-02-05
Dataset sample and donor endpoints

ChuckKollar:
Endpoint to return 'samples', 'organs', and 'donors' uuids of dataset id given.

Issue: https://github.com/hubmapconsortium/entity-api/issues/604

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2626,8 +2626,8 @@ def get_associated_organs_from_dataset(id):
 
     return jsonify(final_result)
 
-@app.route('/datasets/<id>/organs-donors-samples-uuids', methods=['GET'])
-def get_associated_organs_donors_samples_uuids_from_dataset(id):
+@app.route('/datasets/<id>/samples', methods=['GET'])
+def get_associated_samples_from_dataset(id):
     # Token is not required, but if an invalid token provided,
     # we need to tell the client with a 401 error
     validate_token_if_auth_header_exists(request)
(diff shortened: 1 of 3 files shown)


src/app_neo4j_queries.py
```diff
@@ -670,6 +670,37 @@ def get_associated_organs_from_dataset(neo4j_driver, dataset_uuid):
...
+        "MATCH (ds:Dataset)<-[*]-(d:Donor) " \
+        f"WHERE ds.uuid='{dataset_uuid}' " \
+        "RETURN DISTINCT d.uuid"
+
+    results: dict = {}
+    with neo4j_driver.session() as session:
+        results['samples'] = session.read_transaction(schema_neo4j_queries.execute_readonly_tx, sample_query)
+        results['organs'] = session.read_transaction(schema_neo4j_queries.execute_readonly_tx, organ_query)
+        results['donors'] = session.read_transaction(schema_neo4j_queries.execute_readonly_tx, donor_query)
+    return results
```
yuanzhou, line 701: Will need to loop through the resulting records for each category, otherwise only one uuid gets added to each array. 

NO NEED to fix since we are moving to a simpler and more efficient solution.

yuanzhou (changes requested):
@ChuckKollar here is the summary of the solution change: https://github.com/hubmapconsortium/ingest-api/issues/411#issuecomment-1932314600

src/app.py
```diff
@@ -2626,6 +2626,83 @@ def get_associated_organs_from_dataset(id):
...
+        token = get_user_token(request, non_public_access_required=True)
+
+    # By now, either the entity is public accessible or
+    # the user token has the correct access level
+    associated_samples = app_neo4j_queries.get_associated_samples_from_dataset(neo4j_driver_instance, entity_dict['uuid'])
+
+    # If there are zero items in the list associated organs, then there are no associated
+    # Organs and a 404 will be returned.
+    if len(associated_samples) < 1:
+        not_found_error("the dataset does not have any associated organs")
```
yuanzhou, line 2659: Should say "samples" instead of "organs" in the 404 message, also update the comment above

src/app.py
```diff
@@ -2626,6 +2626,83 @@ def get_associated_organs_from_dataset(id):
...
+        token = get_user_token(request, non_public_access_required=True)
+
+    # By now, either the entity is public accessible or
+    # the user token has the correct access level
+    associated_donors = app_neo4j_queries.get_associated_donors_from_dataset(neo4j_driver_instance, entity_dict['uuid'])
+
+    # If there are zero items in the list associated organs, then there are no associated
+    # Organs and a 404 will be returned.
+    if len(associated_donors) < 1:
+        not_found_error("the dataset does not have any associated organs")
```
yuanzhou, line 2698: Fix error message and comment above line to say "donors".

src/app_neo4j_queries.py
```diff
@@ -670,6 +670,44 @@ def get_associated_organs_from_dataset(neo4j_driver, dataset_uuid):
 
     return results
 
+def get_associated_samples_from_dataset(neo4j_driver, dataset_uuid):
```
yuanzhou, line 673: Please add the comment block above the function declaration

src/app_neo4j_queries.py
```diff
@@ -670,6 +670,44 @@ def get_associated_organs_from_dataset(neo4j_driver, dataset_uuid):
...
+
+    with neo4j_driver.session() as session:
+        record = session.read_transaction(schema_neo4j_queries.execute_readonly_tx, query)
+
+        if record and record[record_field_name]:
+            results = schema_neo4j_queries.nodes_to_dicts(record[record_field_name])
+
+    return results
+
+def get_associated_donors_from_dataset(neo4j_driver, dataset_uuid):
```
yuanzhou, line 692: Please add the comment block above the function declaration

yuanzhou (changes requested):

src/app.py
ChuckKollar: OK!

src/app_neo4j_queries.py
ChuckKollar: OK!

src/app_neo4j_queries.py
ChuckKollar: OK!

yuanzhou (approved):

PR #616, opened 2024-02-07
Generalized sublassing. Publications in the provenance schema yaml no…

DerekFurstPitt:
Generalized sublassing. Publications in the provenance schema yaml now only includes properties that are different than datasets. The rest it inherits using yaml anchors and aliases. In schema_manager generate_triggered_data, We now need to remove the null values from properties that are None.

### DIFF
--- a/src/schema/provenance_schema.yaml
+++ b/src/schema/provenance_schema.yaml
@@ -270,7 +270,7 @@ ENTITIES:
     derivation:
       source: true
       target: true
-    properties:
+    properties: &shared_dataset_properties
       <<: *shared_properties
       <<: *shared_entity_properties
       <<: *doi_properties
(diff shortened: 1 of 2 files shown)


src/schema/provenance_schema.yaml
```diff
@@ -270,7 +270,7 @@ ENTITIES:
     derivation:
       source: true
       target: true
-    properties:
+    properties: &shared_dataset_properties
       <<: *shared_properties
       <<: *shared_entity_properties
       <<: *doi_properties
```
yuanzhou, line 276: Since the `&shared_dataset_properties` already contains other shared properties, no need to specify them again:
```
<<: *shared_properties
<<: *shared_entity_properties
<<: *doi_properties
```

src/schema/schema_manager.py
```diff
@@ -288,6 +288,8 @@ def generate_triggered_data(trigger_type, normalized_class, user_token, existing
     # The ordering of properties of this entity class defined in the yaml schema
     # decides the ordering of which trigger method gets to run first
     properties = schema_section[normalized_class]['properties']
+    # Remove null properties such as those not inherited from a superclass
+    properties = remove_none_values(properties)
```
yuanzhou, line 292: Removing the None properties (the ones considered as "private" in Superclass) here only takes care of the triggered data. It doesn't take schema validation into account, which is still used during the creation POST and update PUT calls to `validate_json_data_against_schema()`. And this will cause errors. In addition, consider an extreme use case that none of the properties have a trigger method, so such removal will never happen.

Instead, the best place to get rid of the None properties is in `load_provenance_schema()`, where we generate a final version of all the properties without those "private" Superclass properties (`Publication.ingest_task` for example). As a result, everywhere else in the schema we'll never need to worry about such Nones.

yuanzhou (changes requested):

src/schema/provenance_schema.yaml
DerekFurstPitt: Good catch, yes this would be redundant

src/schema/schema_manager.py
DerekFurstPitt: I see. My primary concern was making sure that the remove_none_values function got run for each entity type at the correct level within the dictionary. Its better to remove right at the beginning, even if it means an extra step to get to the correct level. Fixed

yuanzhou (approved):

PR #628, opened 2024-02-27
Derek furst/hubma id in instanceof

DerekFurstPitt:
converted hubmapid to uuid before running schema_manager.entity_instanceof()

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -473,6 +473,14 @@ def get_entities_instanceof(id, type):
     try:
         uuid = schema_manager.get_hubmap_ids(id.strip())['uuid']
         instanceof: bool = schema_manager.entity_instanceof(uuid, type)
+    except requests.exceptions.RequestException as e:
+        status_code = e.response.status_code
+        if status_code == 400:
+            bad_request_error(e.response.text)
+        if status_code == 404:
+            not_found_error(e.response.text)
+        else:
+            internal_server_error(e.response.text)
     except:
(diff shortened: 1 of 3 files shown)


src/app.py
```diff
@@ -471,7 +471,8 @@ def get_ancestor_organs(id):
 @app.route('/entities/<id>/instanceof/<type>', methods=['GET'])
 def get_entities_instanceof(id, type):
     try:
-        instanceof: bool = schema_manager.entity_instanceof(id, type)
+        uuid = schema_manager.get_hubmap_ids(id.strip())['uuid']
+        instanceof: bool = schema_manager.entity_instanceof(uuid, type)
     except:
```
yuanzhou, line 476: @DerekFurstPitt the issue with your change by using `get_hubmap_ids(id.strip())['uuid']` is when the target id is not found by uuid-api (404 error) or uuid-api considers the id format invalid (404 error), the client will always get "Unable to process request" 400 error due to the `try/except` handling. 

Let's handle such exceptions particularly (still keep the generic `except` as what we have now as the last resort) to create a better user experience with more precise error messages. You can reuse the code from https://github.com/hubmapconsortium/entity-api/blob/51949247876c8c84077a6ce62e8d3b5933d4ba79/src/app.py#L4809-L4819

yuanzhou (changes requested):

yuanzhou (approved):

Issue #635, opened 2024-03-05
Add attributes to track existing component datasets after conversion to Multi-Assay


shirey:
Add two new attributes to Dataset which will be used to track Multi-Assay datasets that are constructed from existing data:
  - `superseded_associated_processed_component_uuids` array of strings holding the uuids of older Datasets that were used to construct a new Multi-Assay dataset, this will be set on new Multi-Assay datasets
  - `new_associated_multi_assay_uuid` a single string holding the uuid of a new Multi-Assay dataset which was, in part, created using an old/existing dataset.  This will be set on the old, existing component datasets which are "reorganized" into new Multi-Assay datasets.

These new attributes will allow tracking of old existing, processed datasets that are associated with newly created Multi-Assay datasets.  Instead of creating a link/edge in Neo4j we will just store ids in the associated datasets.  This will allow us to track the links shown in yellow in the drawing at the top of [Multi-Assay Reorg](https://docs.google.com/document/d/1e6a2HJzbkLkOHj55bLARWBKaqTAuRxQLd_6P8-CRXPI/edit?usp=sharing)

![image](https://github.com/hubmapconsortium/entity-api/assets/5816400/f643baae-3c11-4573-bf2e-68cbdac910bb)

kburke:
@shirey Should trigger methods do the following?

- Verify no Dataset in the `superseded_associated_processed_component_uuids` array is itself a Multi-Assay Dataset, by verifying none have their own `superseded_associated_processed_component_uuids` array?
- Verify no Dataset in the `superseded_associated_processed_component_uuids` array is already a part of another Multi-Assay Dataset, by verifying none have their `new_associated_multi_assay_uuid` set?

Also, Multi-Assay Datasets can only be POSTed, with all of the component Datasets specified?  There is no notion of PUTting later, particularly to add or delete component Datasets?

kburke:
Per Bill's feedback, I will make sure the Dataset POSTed does not specify both `superseded_associated_processed_component_uuids` and `new_associated_multi_assay_uuid`, but will not navigate submitted UUIDs for deeper verification at this time.

I will limit setting `superseded_associated_processed_component_uuids` to POST call which create a Multi-Assay Dataset, and reject any PUT that modifies it.

POST and PUT calls which set `new_associated_multi_assay_uuid` will only be allowed when  `new_associated_multi_assay_uuid` and `superseded_associated_processed_component_uuids` are not set.

PR #639, opened 2024-03-12
Initial commit of changes supporting Multi-Assay Dataset tracking

kburke:
Initial commit of changes supporting Multi-Assay Dataset tracking for entity-api #635. Added superseded_associated_processed_component_uuids and new_associated_multi_assay_uuid attributes to Datasets.

### DIFF
--- a/src/schema/schema_validators.py
+++ b/src/schema/schema_validators.py
@@ -46,7 +46,6 @@ def validate_application_header_before_entity_create(normalized_entity_type, req
 
 
 """
-@TODO-KBKBKB redo doc...
 Validate the specified value for a Dataset's dataset_type is in the valueset UBKG recognizes. 
 
 Parameters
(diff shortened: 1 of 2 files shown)


yuanzhou (changes requested):
@kburke, we'll need to make two additional changes:

- Check the existence of the provided uuid from each property upfront. Can be handled in a similar fashion: https://github.com/hubmapconsortium/entity-api/blob/karlburke/AddAttribsToTrackMultiAssayComponents/src/app.py#L1243-L1264 

- These two new Dataset properties should not be available to Publication. Because Publication is considered as a subclass of Dataset, they'll be inherited by Publication unless you set them to `null` in the yaml. Example: https://github.com/hubmapconsortium/entity-api/blob/karlburke/AddAttribsToTrackMultiAssayComponents/src/schema/provenance_schema.yaml#L633

src/schema/schema_validators.py
```diff
@@ -626,6 +625,73 @@ def validate_group_name(property_key, normalized_entity_type, request, existing_
...
+    # If no contradictions above have caused a ValueError, check if new data contains UUIDs for valid entities.
+    if  'new_associated_multi_assay_uuid' in new_data_dict:
+        proposedMultiAssayDataset = schema_neo4j_queries.get_entity(schema_manager.get_neo4j_driver_instance()
+                                                                    , new_data_dict['new_associated_multi_assay_uuid'])
+        if len(proposedMultiAssayDataset) < 1:
+            raise ValueError(   f"'new_associated_multi_assay_uuid' value"
+                                f" {new_data_dict['new_associated_multi_assay_uuid']}"
+                                f" does not exist.")
+    if  'superseded_associated_processed_component_uuids' in new_data_dict:
+        proposedComponentsList = schema_manager.convert_str_literal(new_data_dict['superseded_associated_processed_component_uuids'])
```
yuanzhou, line 685: We don't need to use `convert_str_literal()` here since the uuids list is already a Python list from converted from request JSON payload.

yuanzhou (changes requested):

yuanzhou (approved):

PR #646, opened 2024-03-22
Derek furst/sync component datasets

DerekFurstPitt:
Changed the status property in provenance schema yaml. Now, its after update trigger method is update_status, which calls set status history, and then calls a new function sync_component_dataset_status. this works by creating new PUT calls to entity on each of the component datasets, updating their status to whatever the multi-assay dataset was. A new validator was also created that disallows directly updating the status of component datasets.

### DIFF
--- a/src/instance/app.cfg.example
+++ b/src/instance/app.cfg.example
@@ -32,6 +32,9 @@ INGEST_API_URL = 'https://ingest-api.dev.hubmapconsortium.org'
 # Works regardless of the trailing slash
 ONTOLOGY_API_URL = 'https://ontology-api.dev.hubmapconsortium.org'
 
+# URL for talking to Entity API (default for DEV)
+ENTITY_API_URL = 'https://entity-api.dev.hubmapconsortium.org'
+
 # A list of URLs for talking to multiple Search API instances (default value used for docker deployment, no token needed)
 # Works regardless of the trailing slash /
 SEARCH_API_URL_LIST = ['http://search-api:8080']
--- a/src/schema/schema_manager.py
+++ b/src/schema/schema_manager.py
@@ -534,7 +534,7 @@ token : string
 def update_entity(uuid, data, token):
     url = _entity_api_url + SchemaConstants.ENTITY_API_UPDATE_ENDPOINT +  '/' + uuid
     header = _create_request_headers(token)
-    header['X-Hubmap-Application'] = 'ingest-api'
+    header[schemaConstants.HUBMAP_APP_HEADER] = INGEST_API_APP
     response = requests.put(url=url, headers=header, json=data)
 """
 Generate the complete entity record by running the read triggers
--- a/src/schema/schema_manager.py
+++ b/src/schema/schema_manager.py
@@ -1,5 +1,4 @@
 import ast
-import sys
 import yaml
 import logging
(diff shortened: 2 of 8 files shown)


src/schema/schema_manager.py
```diff
@@ -507,6 +516,25 @@ def remove_transient_and_none_values(merged_dict, normalized_entity_type):
...
+    The uuid of the entity being updated
+data : dict
+    A dict representation of the json_data_dict for the updated entity
+token : string
+    The globus groups token
+"""
+def update_entity(uuid, data, token):
+    url = _entity_api_url + SchemaConstants.ENTITY_API_UPDATE_ENDPOINT +  '/' + uuid
+    header = _create_request_headers(token)
+    header[schemaConstants.HUBMAP_APP_HEADER] = INGEST_API_APP
```
yuanzhou, line 536: I didn't test, at a quick glance `INGEST_API_APP` should be referenced using `schemaConstants.`.

src/schema/schema_manager.py
```diff
@@ -507,6 +516,25 @@ def remove_transient_and_none_values(merged_dict, normalized_entity_type):
...
+Parameters
+----------
+uuid : string
+    The uuid of the entity being updated
+data : dict
+    A dict representation of the json_data_dict for the updated entity
+token : string
+    The globus groups token
+"""
+def update_entity(uuid, data, token):
```
yuanzhou, line 533: Instead of defining this `update_status()` in schema_manager, we should handle the business logic in the trigger method. Because the schema_manager should not care about updating the entity specifically. 

We should add a getter method similar to https://github.com/hubmapconsortium/entity-api/blob/795b63d92eeee114aac28583c3bbae882e41adb2/src/schema/schema_manager.py#L1690-L1701 
so other modules can get a handle of the entity-api url and then make request inside the trigger method like this https://github.com/hubmapconsortium/entity-api/blob/795b63d92eeee114aac28583c3bbae882e41adb2/src/schema/schema_triggers.py#L1390

yuanzhou (changes requested):

src/instance/app.cfg.example
```diff
@@ -32,6 +32,9 @@ INGEST_API_URL = 'https://ingest-api.dev.hubmapconsortium.org'
 # Works regardless of the trailing slash
 ONTOLOGY_API_URL = 'https://ontology-api.dev.hubmapconsortium.org'
 
+# URL for talking to Entity API (default for DEV)
+ENTITY_API_URL = 'https://entity-api.dev.hubmapconsortium.org'
```
yuanzhou, line 36: Even though this is instructional, we should use localhost URL rather than DEV. Add some comment to explain this is the same URL base where the entity-api is running. Because if someone uses the DEV URL and runs the entity-api locally, it'll end up updating the component datasets in DEV neo4j from the local trigger requests. The key is to keep it consistent. And once deployed on DEV/TEST/PROD, we can still use a `localhost:port` where the entity-api runs on the VM.

yuanzhou (changes requested):

DerekFurstPitt:
@yuanzhou I applied the requested fixes. Take a look when you get a chance

yuanzhou (approved):

Issue #630, opened 2024-03-05
New features oriented to improve index procedure efficiency


yuanzhou:
During indexing, all we need are the node properties from Neo4j directly AND a few `on_read_trigger` generated ones. But the current `GET /entitites/<id>` returns ALL trigger generated properties, totally inefficient and unnecessary. And the index procedure has to remove the ones that are not specified in the mapping json with lots of `for` loops, such repetitive work is a total waste of time and further reduces the efficiency. 

To address the above issues in the entity-api, 

- Introduce a new trigger type: `on_index_trigger`, which has the same trigger method as regular `on_read_trigger`.
- Add a specialized schema_manager method `get_complete_document_result()` which is very similar to `get_complete_entity_result()` (for now do NOT modify this method), but uses `on_index_trigger` instead.
- For caching, use a different prefix on those index data, `cache_key = f'{_memcached_prefix}_complete_index_{entity_uuid}'`
- Introduce a new flag in schema yaml `indexed: false` (default is true) for indexing purposes. This allows us to remove the use of json mapping in search-api. When `exposed: false` there's no need to check this `indexed: false` since it's a field won't be exposed by entity-api. When there's only `indexed: false` it means this field still gets returned to the regular GET call, but we won't index this field.
- Add a specialized schema_manager method `normalize_document_result_for_response()` based on the current `normalize_entity_result_for_response()` and integrate with the `indexed` flag.

Afterwards, create a specialized endpoint (based on the current `GET /entities/<id>` but with no property filtering needed): `GET /documents/<id>` which returns a subset of the regular entity json to be used for index, without including the following generated fields (not used by index process):

- Donor: None, since no on_read_trigger
- Sample: `direct_ancestor`
- Dataset/Publication (including revisions): `collections`, `upload`, `direct_ancestors`, `local_directory_rel_path`
- Publication-specific: ~~`associated_collection`~~ (still need this field to be indexed)

**AFTER** the new features get tested, switch to use this new endpoint in search-api: https://github.com/hubmapconsortium/search-api/issues/756

kburke:
As I understand directions in the Description, I believe `local_directory_rel_path` should not be a part of the OpenSearch document.

[...]

PR #649, opened 2024-04-01
Initial commit for entity-api #630.

kburke:
Initial commit for entity-api #630.
Added on_index_trigger and supporting code to efficiently create documents for OpenSearch indices.  Also some typo corrections, reformatting, and swapping some string comparison for an enum.


src/schema/provenance_schema.yaml
```diff
@@ -436,27 +443,31 @@ ENTITIES:
...
         on_read_trigger: get_previous_revision_uuid
+        # No on_index_trigger to include previous_revision_uuid in the OpenSearch document for a Dataset
       previous_revision_uuids:
         type: list
         generated: true
         transient: true
         immutable: true
         description: "The list of the uuids of previous revision datasets"
         on_read_trigger: get_previous_revision_uuids
+        # No on_index_trigger to include previous_revision_uuids in the OpenSearch document for a Dataset
```
yuanzhou, line 454: We still need to index this field

src/schema/schema_manager.py
```diff
@@ -697,6 +838,104 @@ def normalize_entity_result_for_response(entity_dict, properties_to_exclude = []
...
+        # Only return the properties defined in the schema yaml
+        # Exclude additional properties if specified
+        if  key in properties and \
+            key in properties_to_skip:
+            # Skip properties if directed by the calling function
+            continue
+        if  entity_dict[key] is None:
+            # Do not include properties in the metadata if they are empty
+            continue
+        if  'exposed' in properties[key] and \
```
yuanzhou, line 909: This will cause `KeyError` issues when a key returned by the neo4j query is not defined in the schema yaml, and we have such fields. There should be an outer "if check" to restrict the scope. 

For example, line 899 can use 
```
if key in entity_dict and key in properties:
```

yuanzhou (changes requested):
@kburke your effort on generalizing the functionalities to accommodate future needs as well as introducing the trigger type constants to replace the hardcoded values are greatly appreciated! Also thanks for fixing the typos. 

I noticed couple of issues in the code, haven't got time to test extensively.

src/schema/provenance_schema.yaml
```diff
@@ -436,27 +443,31 @@ ENTITIES:
...
       next_revision_uuid:
         type: string
         generated: true
         transient: true
         immutable: true
         description: "The uuid of next revision dataset"
         on_read_trigger: get_next_revision_uuid
+        # No on_index_trigger to include next_revision_uuid in the OpenSearch document for a Dataset
```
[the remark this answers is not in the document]
kburke: Fixed in https://github.com/hubmapconsortium/entity-api/commit/205e36ea8ff7b6fb744f3dc5276fdb62a1e8affb

yuanzhou (changes requested):

yuanzhou (approved):

PR #660, opened 2024-04-15
Initial revisions for entity-api/issues/631.

kburke:
Initial revisions for entity-api/issues/631. Move calculation of display_subtype from search-api to entity-api to take advantage of caching for indexing efficiency. Companion revision in search-api removes display_subtype.


src/schema/schema_manager.py
```diff
@@ -40,6 +40,7 @@
 _memcached_client = None
 _memcached_prefix = None
 
+ONTOLOGY_API_ORGAN_TYPES_ENDPOINT = '/organs/by-code?application_context=HUBMAP'
```
yuanzhou, line 43: Duplicate of https://github.com/hubmapconsortium/entity-api/blob/5c3f411f58f8a9cd6bf75ef8994c31cfa4f9bf0a/src/schema/schema_constants.py#L21

src/schema/provenance_schema.yaml
```diff
@@ -594,6 +600,12 @@ ENTITIES:
         generated: true
         immutable: true
         description: "The assay types of this Dataset. Valid values are from UBKG are queried by schema_manager.get_valueset_dataset_type() using the Ontology API."
+      display_subtype:
```
yuanzhou, line 603: Publication is considered as a subclass of Dataset, since `display_subtype` is defined for Dataset, no need to redefine for Publication.

yuanzhou (changes requested):

yuanzhou (approved):

PR #663, opened 2024-04-22
added constraint to /components endpoint

DerekFurstPitt:
Added constraint to /components endpoint. If any of the direct ancestors provided already has component children, Bad_request_error

### DIFF
--- a/src/app_neo4j_queries.py
+++ b/src/app_neo4j_queries.py
@@ -114,16 +114,13 @@ Returns
 boolean 
 """
 def dataset_has_component_children(neo4j_driver, dataset_uuid):
-    query = (f"MATCH (ds1:Dataset)<-[:ACTIVITY_OUTPUT]-(a:Activity)<-[:ACTIVITY_INPUT]-(ds2:Dataset) "
-             f"WHERE ds2.uuid = '{dataset_uuid}' "
-             f"RETURN a.creation_action as {record_field_name}")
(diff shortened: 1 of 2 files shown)


src/app_neo4j_queries.py
```diff
@@ -97,6 +97,34 @@ def get_entities_by_type(neo4j_driver, entity_type, property_key = None):
...
+    query = (f"MATCH (ds1:Dataset)<-[:ACTIVITY_OUTPUT]-(a:Activity)<-[:ACTIVITY_INPUT]-(ds2:Dataset) "
+             f"WHERE ds2.uuid = '{dataset_uuid}' "
+             f"RETURN a.creation_action as {record_field_name}")
+    with neo4j_driver.session() as session:
+        components_list = session.run(query).data()
+    for component in components_list:
+        creation_action = component.get('result')
+        if creation_action == 'Multi-Assay Split':
+            return True
+    return False
```
yuanzhou, line 126: Can use a much simpler check to return the boolean directly

```
MATCH p=(ds1:Dataset)<-[:ACTIVITY_OUTPUT]-(a:Activity)<-[:ACTIVITY_INPUT]-(ds2:Dataset) 
WHERE ds2.uuid = '{dataset_uuid}' AND a.creation_action = "Multi-Assay Split"
RETURN (count(p) > 0) as {record_field_name}
```

yuanzhou (changes requested):

yuanzhou (approved):

PR #664, opened 2024-04-24
Remove search-api's neo4j-to-es-attributes.json and replace with indexed: indicators on attributes in entity-api's provenance_schema.yaml

kburke:
Initial changes for search-api/issues/753. Remove search-api's neo4j-to-es-attributes.json and replace with indexed: indicators on attributes in entity-api's provenance_schema.yaml.

### DIFF
--- a/src/schema/provenance_schema.yaml
+++ b/src/schema/provenance_schema.yaml
@@ -94,7 +94,7 @@ ACTIVITIES:
       <<: *shared_properties
       creation_action:
         type: string
-        indexed: true
+        indexed: false
         description: "The activity that was performed."
         before_create_trigger: set_activity_creation_action
 
@@ -503,7 +503,7 @@ ENTITIES:
         generated: true
         transient: true
         immutable: true
-        indexed: true
+        indexed: false
         description: "The list of the uuids of previous revision datasets"
         on_read_trigger: get_previous_revision_uuids
         # No on_index_trigger to include previous_revision_uuids in the OpenSearch document for a Dataset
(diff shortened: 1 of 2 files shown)


src/schema/provenance_schema.yaml
```diff
@@ -453,6 +503,7 @@ ENTITIES:
         generated: true
         transient: true
         immutable: true
+        indexed: true
```
yuanzhou, line 506: Do not index `previous_revision_uuids`

src/schema/provenance_schema.yaml
```diff
@@ -469,6 +521,7 @@ ENTITIES:
...
```
yuanzhou, line 524: Do not index `next_revision_uuids`

src/schema/provenance_schema.yaml
```diff
@@ -88,6 +94,7 @@ ACTIVITIES:
       <<: *shared_properties
       creation_action:
         type: string
+        indexed: true
```
yuanzhou, line 97: Out index procedure doesn't handle Activity nodes. This `creation_action` from Dataset and Publication does get indexed.

yuanzhou (changes requested):

src/schema/provenance_schema.yaml
kburke: Pushed these corrections.  I assume comment above should say, "... `creation_action` ... does **_NOT_** get indexed."

yuanzhou (approved):

PR #671, opened 2024-04-30
Disabled multiple revisions

DerekFurstPitt:
Removed the ability to include multiple previous revision id's when creating or updating an entity. 

Added validator that sees if a provided previous revision already has a revision itself and raises a value error if it does. 

Still allow previous_revision_uuid value to be a list or a string, if its a list, its just constrained to a single entry for the sake of compatibility. 

Left any underlying schema queries or other methods intact in case we need to use them later, just removed their uses in app.py and the provenance schema. 

Commented out the endpoint GET multi-revisions rather than deleting in case we need to add it back later.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1028,11 +1028,9 @@ def create_entity(entity_type):
         # Also check existence of the previous revision dataset if specified
         if 'previous_revision_uuid' in json_data_dict:
             if isinstance(json_data_dict['previous_revision_uuid'], list):
+                if len(json_data_dict['previous_revision_uuid']) > 1:
+                    bad_request_error(f"Multiple previous revision uuids were provided. Only one is allowed")
                 previous_revision_list = json_data_dict['previous_revision_uuid']
-
-                nested_revisions =  app_neo4j_queries.nested_previous_revisions(neo4j_driver_instance, previous_revision_list)
-                if nested_revisions:
-                    bad_request_error(f"{nested_revisions[0][0]} is a revision of {nested_revisions[1][0]}. Datasets in previous_revision_uuid must not be revisions of eachother")
             else:
                 previous_revision_list = [json_data_dict['previous_revision_uuid']]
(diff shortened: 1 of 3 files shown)


src/app.py
```diff
@@ -1029,11 +1029,9 @@ def create_entity(entity_type):
         # Also check existence of the previous revision dataset if specified
         if 'previous_revision_uuid' in json_data_dict:
             if isinstance(json_data_dict['previous_revision_uuid'], list):
+                if len(json_data_dict['previous_revision_uuid']) > 1:
```
yuanzhou, line 1032: This `previous_revision_uuid` should always be a string, there's really no point to treat it as a list. Looking back I found this commit: https://github.com/hubmapconsortium/entity-api/commit/f37aa8dbc5c78d2c89bdc436d23aedfb39aa357a#diff-cd8a3bff2418f86ca995d983b9e0d2ed942f3a2a10b8eb393b15f18f56ea3fb4R421-R423 

We should remove the `- list` from yaml.

src/schema/schema_validators.py
```diff
@@ -127,6 +127,23 @@ def validate_dataset_not_component(property_key, normalized_entity_type, request
...
 
 
+"""
+If the provided previous revision is already a revision of another dataset, disallow
+"""
+def validate_if_revision_is_unique(property_key, normalized_entity_type, request, existing_data_dict, new_data_dict):
+    previous_revision = new_data_dict['previous_revision_uuid']
+    neo4j_driver_instance = schema_manager.get_neo4j_driver_instance()
+    if isinstance(previous_revision, list):
+        previous_revision_uuid = previous_revision[0]
```
yuanzhou, line 137: Since we'll only define this field as a string, no need to check it as a list.

yuanzhou (changes requested):

yuanzhou (approved):

PR #686, opened 2024-06-11
added endpoints entities/<id>/collections

DerekFurstPitt:
added GET endpoints /entities/<id>/collections and /entities/<id>/uploads and accompanying schema query methods.


yuanzhou (approved):
Will test on DEV

PR #690, opened 2024-06-20
Derek furst/external process

DerekFurstPitt:
Modified validator to allow creation of new creation_action "external process"


yuanzhou (changes requested):
@DerekFurstPitt there are two things we'll need to verify and handle properly:
- The direct ancestors of this `External Process` generated dataset can only be Dataset (not Sample or Donor)
- Publication can not be created with this new action. Currently Publication is a subclass of Dataset, and `creation_action` can still be provided in the POST and PUT. It appears to me that we never really handled the case the Publication to prevent this `creation_action` being used.

src/schema/schema_neo4j_queries.py
```diff
@@ -112,6 +112,37 @@ def get_entity(neo4j_driver, uuid):
...
+"""
+Get the uuids for each entity in a list that doesn't belong to a certain entity type. Uuids are ordered by type
+
+Parameters
+----------
+neo4j_driver : neo4j.Driver object
+    The neo4j database connection pool
+direct_ancestor_uuids : list
+    List of the uuids to be filtered
+
```
yuanzhou, line 125: Missing `entity_type` in parameters list

src/schema/schema_validators.py
```diff
@@ -579,13 +579,21 @@ def validate_publication_date(property_key, normalized_entity_type, request, exi
...
-    accepted_creation_action_values = ["central process", "lab process"]
+    accepted_creation_action_values = ["central process", "lab process", "external process"]
     creation_action = new_data_dict[property_key].lower()
     if creation_action and creation_action not in accepted_creation_action_values:
         raise ValueError("Invalid {} value. Accepted values are: {}".format(property_key, ", ".join(accepted_creation_action_values)))
     if creation_action == '':
         raise ValueError(f"The property {property_key} cannot be empty, when specified.")
-
+    if creation_action == 'external process':
+        if normalized_entity_type.lower() != "dataset":
```
yuanzhou, line 589: This validator is specific to Dataset, and Publication already rejects the `creation_action` in request payload at the schema level. This check becomes unnecessary and will never satisfy the condition.

src/schema/schema_neo4j_queries.py
```diff
@@ -112,6 +112,37 @@ def get_entity(neo4j_driver, uuid):
...
+Returns
+-------
+dict
+    A dictionary of entity uuids that don't pass the filter, grouped by entity_type
+"""
+def filter_ancestors_by_type(neo4j_driver, direct_ancestor_uuids, entity_type):
+    uuids_str = ", ".join([f'"{uuid}"' for uuid in direct_ancestor_uuids])
+    query = (f"MATCH (e:Entity) "
+             f"WHERE e.uuid in [{uuids_str}] AND toLower(e.entity_type) <> '{entity_type.lower()}' "
```
yuanzhou, line 134: You can use `WHERE e.uuid in {direct_ancestor_uuids}` directly without joining the uuids into `uuids_str`, can be more efficient.

yuanzhou (changes requested):
@DerekFurstPitt your updates look good and work well, we can use some minor tweaks to make the code more efficient.

yuanzhou (approved):

PR #706, opened 2024-07-29
Derek furst/constraints

DerekFurstPitt:
Implemented constraints endpoint in hubmap with the only constraint being:  only samples of sample_category "Organ" are allowed to be descendants of donors.


yuanzhou (changes requested):
@DerekFurstPitt a few changes per our talk and Gesina's response:
- Use a single Python module `metadata_constraints.py` replace the `lib/constraints` package
- Use constants or just a Python list in the above module to define allowed entity types, allowed ancestors and descendants to be used by the constraints. No need to make it configurable.
- Add the latest DeepDiff 7.0.1 to requirements.txt
- Remove the `filter` and `report_type` query parameters, not needed
- Remove unnecessary wrapper functions and unused function
- Prevent 500 error when a function is not defined but being called, can use some existence checks based on the constants defined earlier.

src/app.py
```diff
@@ -115,6 +120,32 @@ def http_internal_server_error(e):
     return jsonify(error = str(e)), 500
 
 
+# ####################################################################################################
+# ## UBKG Ontology and REST initialization
+# ####################################################################################################
```
yuanzhou, line 125: @DerekFurstPitt let's remove the commented code since they are not being used

src/app.py
```diff
@@ -2256,6 +2287,75 @@ def get_uploads(id):
...
+"""
+@app.route('/constraints', methods=['POST'])
+def validate_constraints():
+    if not request.is_json:
+        bad_request_error("A json body and appropriate Content-Type header are required")
+    json_entry = request.get_json()
+    is_match = request.values.get('match')
+    order = request.values.get('order')
+    use_case = request.values.get('filter')
+    report_type = request.values.get('report_type')
```
yuanzhou, line 2328: I thought you had remove `filter` and `report_type` and their handling below...

src/instance/app.cfg.example
```diff
@@ -64,3 +64,5 @@ DOI_REDIRECT_URL = 'https://portal.hubmapconsortium.org/browse/<entity_type>/<id
 # OLD: https://raw.githubusercontent.com/hubmapconsortium/ccf-releases/main/reference-entity-ids.tsv
 # New URL in use since 7/2/2024 following the HRA 2.1 release
 REDIRECTION_INFO_URL = 'https://cdn.humanatlas.io/digital-objects/reference-entity-ids.tsv'
+
+ENTITY_TYPES = ['dataset', 'sample', 'donor', 'publication', 'upload']
```
yuanzhou, line 68: Remove unwanted config

yuanzhou (changes requested):
@DerekFurstPitt we'll also need to delete the `lib/constraints`

yuanzhou (changes requested):

yuanzhou (approved):

PR #712, opened 2024-08-02
added validator for sample.direct_ancestor_uuid and datasets.direct_ancestor_uuids which prevents an entity's own uuid from being included in direc_ancestor_uuid(s) during a put call

DerekFurstPitt:


src/schema/provenance_schema.yaml
```diff
@@ -428,6 +428,7 @@ ENTITIES:
         before_property_update_validators:
           - validate_no_duplicates_in_list
           - validate_not_invalid_creation_action
+          - validate_id_not_in_direct_ancestor
```
yuanzhou, line 431: This validator should also be defined for the entity creation.

src/schema/provenance_schema.yaml
```diff
@@ -938,6 +939,8 @@ ENTITIES:
         # Note: link_sample_to_direct_ancestor() will always delete all the old linkages first
         after_create_trigger: link_sample_to_direct_ancestor
         after_update_trigger: link_sample_to_direct_ancestor
+        before_property_update_validators:
+          - validate_id_not_in_direct_ancestor
```
yuanzhou, line 943: Same as above comment

src/schema/schema_validators.py
```diff
@@ -561,6 +561,36 @@ def validate_publication_date(property_key, normalized_entity_type, request, exi
...
+new_data_dict : dict
+    The json data in request body, already after the regular validations
+"""
+def validate_id_not_in_direct_ancestor(property_key, normalized_entity_type, request, existing_data_dict, new_data_dict):
+    if 'uuid' not in existing_data_dict:
+        raise KeyError("Missing 'uuid' key in 'existing_data_dict' during calling 'validate_id_not_in_direct_ancestor()' validator method.")
+    entity_uuid = existing_data_dict.get("uuid")
+    if property_key == 'direct_ancestor_uuid':
+        ancestors = new_data_dict.get('direct_ancestor_uuid')
+    elif property_key == 'direct_ancestor_uuids':
```
yuanzhou, line 588: There's no need to check the name of the `property_key`, as long as this validator is defined in the yaml for the given field, it'll be only for that field name.

yuanzhou (changes requested):

yuanzhou (approved):

PR #718, opened 2024-08-07
Added  schema_validators.validate_application_header_before_property_…

ChuckKollar:
…update to checks for entity_bulk_update()

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -4251,7 +4251,7 @@ def bulk_update_entities(
 ) -> dict:
     headers = {
         "Authorization": f"Bearer {token}",
-        "X-Hubmap-Application": SchemaConstants.ENTITY_API_APP,
+        SchemaConstants.HUBMAP_APP_HEADER: SchemaConstants.ENTITY_API_APP,
     }
     # create a session with retries
     session = requests.Session()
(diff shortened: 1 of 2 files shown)


src/app.py
```diff
@@ -4337,6 +4338,10 @@ def entity_bulk_update():
 
     validate_token_if_auth_header_exists(request)
     require_json(request)
+    try:
+        schema_validators.validate_application_header_before_property_update(request)
```
yuanzhou, line 4342: @ChuckKollar we do not need to explicitly call the validator here. When you make the PUT calls against entity-api itself to update each dataset/upload, it'll go through the yaml validation: https://github.com/hubmapconsortium/entity-api/blob/8b3e20f9e25a8542bacfd768abbdb71f0fcb86c6/src/app.py#L1165

Are you not getting the 400 response without this explicit handling?

yuanzhou (changes requested):

src/app.py
ChuckKollar: When I remove the EXPLCIT CALL I am NOT getting an error when
 --header "X-Hubmap-Application: entity-apix"
 for example.
So, it would appear that the explicit call to the validator needs to be there.

src/schema/schema_validators.py
```diff
@@ -286,23 +286,17 @@ def collection_entities_are_existing_datasets(property_key, normalized_entity_ty
...
-existing_data_dict : dict
-    A dictionary that contains all existing entity properties
-new_data_dict : dict
-    The json data in request body, already after the regular validations
 """
-def validate_application_header_before_property_update(property_key, normalized_entity_type, request, existing_data_dict, new_data_dict):
-    # A list of applications allowed to update this property
-    # Currently only ingest-api and ingest-pipeline are allowed
-    # to update Dataset.status or Upload.status
+def validate_application_header_before_property_update(request):
```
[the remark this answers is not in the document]
ChuckKollar: As Joe requested in slack.... I reverted the change to the signature of the validator. I added status to the payload. It still does not work locally without the explicit call to the validator. But this is what he wanted....

yuanzhou (approved):

PR #725, opened 2024-08-21
Reworked get_organ_types 

DerekFurstPitt:
reworked get_organ_types such that it is only called once during the initialization of schema_manager. Created a global variable inside schema triggers and a quick setter function inside app.py to set that value after initialization of schema manager but before the value is fetched in schema_triggers

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -228,6 +228,8 @@ except Exception:
             f" _schema_yaml_file={_schema_yaml_file}."
     # Log the full stack trace, prepend a line with our message
     logger.exception(msg)
+organ_types_dict = schema_manager.organ_types_dict
+schema_triggers.set_organ_types_dict()
 
 ####################################################################################################
 ## Initialize an S3Worker from hubmap-commons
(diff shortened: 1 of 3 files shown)


yuanzhou (changes requested):
@DerekFurstPitt Your rework looks be an anti-pattern. As you can tell we don't use any top-level variables in `schema_triggers.py`, it contains only functions. And `schema_manager.py` uses global variables but it only exposes them via getter functions, `schema_manager.get_organ_types()` in this case. 

Instead of working around circular import, there's a much simpler solution. Take a look at how `auth_helper_instance = get_auth_helper_instance()` is being used in `schema_manager.py` first. Then
- Add this new `_organ_types = None`
- In `get_organ_types()`, first check `if _organ_types` is available then just return it. Otherwise set its value to be the result of the first response. This is essentially a **singleton pattern**.
- In `schema_triggers.py` and `app.py`, keep using the old `schema_manager.get_organ_types()` since now it just reuses the value.
- In `app.py`, there's no need to use the second parameter for `validate_organ_code(organ_code, organ_types_dict=None)`.

Keep in mind that we are using the `schema_manager` as the central piece, any values to be reused by `app.py` or `schema_triggers.py` just need a reference/pointer to the exposed function in `schema_manager`.

yuanzhou (approved):

PR #729, opened 2024-08-27
Updates to Derek-Furst/constraints

gesinaphillips:
This PR updates previous work on the constraints endpoint to do the following:
- Return a 200 for valid entity types (e.g. dataset, publication, etc.) that do not have constraints.
- Allow an organ to be an ancestor of other entities (except datasets) rather than just a descendant of donor.
- Handle sub_types/sub_type_vals in cases where the constraint has a required sub_type and/or sub_type_val

One uncertainty:
- Different code paths are followed for a request with the URL parameter `order=ancestors` (the default) vs. `order=descendants`. Whichever order is selected (ancestors or descendants), that key in the payload is evaluated first. So if you pass an allowable* entity_type ancestor in an `order=ancestor` request, but that entity_type doesn't have constraints, you will get a 200 back *regardless of whether the descendant is an allowable entity type.* This is probably functionally fine but also could lead to a positive response for the following:
```
curl --location 'http://entity.api.hubmapconsortium.org/constraints?match=True&order=descendants' \
--header 'Authorization: Bearer <token>' \
--header 'Content-Type: application/json' \
--data '[
    {
        "ancestors": [{
            "entity_type": "asdfasdf",
            "sub_type": ["blah"],
            "sub_type_val": "bad"
        }],
        "descendants": [{
            "entity_type": "sample",
            "sub_type": ["block"],
            "sub_type_val": null
        }]
    }
]'
```

```
{
    "code": 200,
    "description": [
        {
            "code": 200,
            "description": "No constraints exist for given descendants: {'entity_type': 'sample', 'sub_type': ['block'], 'sub_type_val': None}",
            "name": "No Constraints"
        }
    ],
    "name": "ok"
}
```

IVT doesn't use `order=descendants` so that could be unsupported for the time being if you don't have plans to use it either. IVT requires the param `order=ancestors` to be valid. This problem solves itself if a comprehensive set of constraints is added down the line.

*allowable meaning part of `entities = ['dataset', 'sample', 'donor', 'publication', 'upload']`

### DIFF
--- a/src/metadata_constraints.py
+++ b/src/metadata_constraints.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional
+from typing import Optional
 from deepdiff import DeepDiff
 
 def build_constraint(ancestor: dict, descendants: list[dict]) -> dict:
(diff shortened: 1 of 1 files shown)


gesinaphillips:
Closing so as not to own this PR

PR #735, opened 2024-09-04
Derek furst/constraints allow non donors

DerekFurstPitt:
Added a bypass inside metadata_constraints.py in the get_constraints function. The original plan was to do the bypass directly in app.py in the endpoint, but because we need to go a few layers down into the request body to find the entity-type, and both that and the part that sets the response code are located there, it worked out better doing the change here. 

now donors will be validated as before (donor must have an organ descendent) but non donors will pass automatically with a "nothing to validate" response like we do elsewhere.


yuanzhou (changes requested):
@DerekFurstPitt a few things:

- Since now you are handling the bypass validation in `metadata_constraints.py`, it makes more sense to also move `constraints_json_is_valid()` to this module so we keep everything in one module.

- Due to `report_type` is no longer needed for HuBMAP, we should [remove the unused `index` variable](https://github.com/hubmapconsortium/entity-api/blob/Derek-Furst/constraints-allow-non-donors/src/app.py#L2332-L2334) too.

- Looks like you are [treating all the non-200 codes as 400 error](https://github.com/hubmapconsortium/entity-api/blob/Derek-Furst/constraints-allow-non-donors/src/app.py#L2341) , then what's the point of returning 404 in `get_constraints()`?

- Within `get_constraints()`, the `result` dict has inconsistent use of `name` and `description`. We can get rid of `name` field and only return `code` and `description` otherwise, only need to handle `name` as part of the endpoint response. Really depends on how you want to handle the final response in `app.py`

yuanzhou (approved):

PR #738, opened 2024-09-13
Derek furst/exclude results

DerekFurstPitt:
implemented field exclusion for public datasets when user not in hubmap read group. Was able to remove sub fields (such as ingest_metadata.metadata.lab_id) through use of a recursive function. Certain endpoints unaffected because they are either not publicly accessible or are otherwise exceptions


yuanzhou (changes requested):
1. `get_fields_to_exclude()`, no need to check Activity section. We haven't had the need to return Activity properties.

2. With or without token, `/documents/<id>` returns the same fields. There's a bug on the access check.

3. Add function doc comments to `get_fields_to_exclude()` and `exclude_properties_from_response()`.

4. Reanme the funtion parameter `json_body` in `exclude_properties_from_response()` because it's a Python dict instead of a json string.

5. dict.copy() used in `exclude_properties_from_response()` is a shallow copy.
It works because it returnd the modified `final_result` directly rather than something else based on the original dict. We should not use shallow copy, but keep in mind copy.deepcopy isn't thread safe, and deepcopy is very slow.

6. In `app.py`, I see different useages of the check, keep consistent:

- if public_entity and not user_in_hubmap_read_group(request):
- if public_entity and user_in_hubmap_read_group(request) is False:
- if public_entity and has_access is False:

7. In `get_entity_by_id()`, `entity_scope = _get_entity_visibility()` is already there, no need to use `public_entity`

PR #780, opened 2025-01-07
updated prov to version 2.0.1. Modified neo4j query to return only ce…

DerekFurstPitt:
…rtain fields for get_provenance, particularly omitting expensive ones like files and metadata


yuanzhou (approved):
@DerekFurstPitt It's great that you used the `CASE` form to construct the returned fields. It worked well during my testing. Two changes I'd like to ask:

1. Add back `hubmap_id` field for Activity nodes. I noticed a rendering issue on portal-ui and reach out to John, he missed this field in his original list. 

2. Constructing the fields programmatically instead of hardcoding them would make the  logic more readable and give us a better control if later we'll add more fields. I put this code snippet as an example:

```
delimiter = ', '
enitty_properties = ['a', 'b', 'c']
cypher_pairs_list = [f'{x}: node.{x}' for x in enitty_properties]
cypher_pairs_str = delimiter.join(cypher_pairs_list) + delimiter
```

DerekFurstPitt:
@yuanzhou requested changes applied

yuanzhou (approved):

Issue #781, opened 2025-01-11
Fully control fields exclusion from public response in schema yaml


yuanzhou:
Following the implementation of #722, we'll also need to exclude those fields nested under the following fields from public responses via the schema yaml:

- `Collection.datasets` (remove the [old implementation](https://github.com/hubmapconsortium/entity-api/pull/745/files) and handle it in schema yaml)
- `Sample.direct_ancestor`
- `Dataset.direct_ancestors` (will be inherited by Publication)

To achieve the above, we'll also need to amend `schema_manager.exclude_properties_from_response()` in a similar fashion as [what SenNet team did](https://github.com/sennetconsortium/entity-api/pull/574/files).

yuanzhou:
For Zhou: these changes will not require any tweaks in search-api, but will need to wait until the schema yaml raw URL cache expires to verify in search-api.

kburke:
Given this work falls in YAML and schema_manager.py files used by the recently added "prov-metadata" endpoint as well as existing code in app.py, I understand this solution will show up throughout entity-api.

I plan to focus my verification on the `/datasets/</id>/prov-metadata` endpoint I've worked on recently.  But please let me know if I should include other endpoints for verification @yuanzhou .

yuanzhou:
@kburke once the changes described in the issue are implemented, we should not see those fields from under `Sample.direct_ancestor` and `Dataset.direct_ancestors` when you call `GET /entities/<id>` without a valid token, regardless if the entity is public or not. And they should become available with a valid token. Also checkout `GET /entities/<collection-id>` and verify the nested fields under `Collection.datasets` list. No need to verify `GET /entities/<upload-id>` in this case since Upload always requires a token to access.

In addition, please also verify the following:

- `/documents/<id>`
- `/entities/<id>/ancestor-organs`
- `/datasets/<id>/donors`
- `/datasets/<id>/organs`
- `/datasets/<id>/samples`

Because we had already skipped `direct_ancestor` and `direct_ancestors` from the following endpoints, so NO need to verify:

- `/ancesotrs/<id>`
- `/descendants/<id>`
- `/parents/<id>`
- `/children/<id>`
- `/entities/<id>/tuplets`
- `/entities/<id>/siblings`

kburke:
[...]

PR #793, opened 2025-01-22
Initial #781 changes, exclude nested fields from public responses

kburke:
Initial changes for https://github.com/hubmapconsortium/entity-api/issues/781 to exclude fields which occur in Collections.datasets, Dataset.direct_ancestors, and Sample.direct_ancestor from public responses.

### DIFF
--- a/src/schema/provenance_schema.yaml
+++ b/src/schema/provenance_schema.yaml
@@ -667,10 +667,6 @@ ENTITIES:
         description: "The activity that was performed."
       dataset_type:
         before_create_trigger: set_publication_dataset_type
-        before_property_create_validators:
-          - validate_recognized_dataset_type
-        before_property_update_validators:
-          - validate_recognized_dataset_type
         type: string
         generated: true
         immutable: true
--- a/src/schema/provenance_schema.yaml
+++ b/src/schema/provenance_schema.yaml
@@ -192,6 +192,10 @@ shared_entity_properties: &shared_entity_properties
 ENTITIES:
   ############################################# Collection #############################################
   Collection:
+    excluded_properties_from_public_response:
+      - datasets:
+        - lab_dataset_id
+        - submission_id
     # Collection can not be derivation source but not target
     derivation:
       source: false
@@ -304,6 +308,12 @@ ENTITIES:
       - lab_dataset_id
       - metadata:
         - lab_id
+      - direct_ancestors:
+        - lab_tissue_sample_id
+        - submission_id
+        - lab_dataset_id
+        - metadata:
+          - lab_id
     derivation:
       source: true
       target: true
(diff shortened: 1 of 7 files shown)


src/schema/provenance_schema.yaml
```diff
@@ -192,6 +192,10 @@ shared_entity_properties: &shared_entity_properties
 ENTITIES:
   ############################################# Collection #############################################
   Collection:
+    excluded_properties_from_public_response:
+      - datasets:
+        - lab_dataset_id
+        - submission_id
```
yuanzhou, line 198: - Remove `submission_id` since dataset doesn't have such field
- Add `metadata.lab_id`, same as https://github.com/hubmapconsortium/entity-api/blob/c367217e7d6355091692072bf4d8cabee75f80b4/src/schema/provenance_schema.yaml#L309-L310

src/schema/provenance_schema.yaml
```diff
@@ -304,6 +308,12 @@ ENTITIES:
       - lab_dataset_id
       - metadata:
         - lab_id
+      - direct_ancestors:
+        - lab_tissue_sample_id
+        - submission_id
+        - lab_dataset_id
+        - metadata:
+          - lab_id
```
yuanzhou, line 316: Considering a Dataset may have multiple parents and they can be Samples or Datasets, for clarity can we add comments to indicate/distinguish which fields belong to what entity type? 

BTW, both Sample and Dataset can have `metadata.lab_id`.

src/schema/provenance_schema.yaml
```diff
@@ -874,6 +880,11 @@ ENTITIES:
     excluded_properties_from_public_response:
       - lab_tissue_sample_id
       - submission_id
+      - direct_ancestor:
+        - lab_donor_id
+        - submission_id
+        - label
+        - lab_tissue_sample_id
```
yuanzhou, line 887: - Similarly, add comments to indicate which fields belong to which entity type for clarity. 
- Add `metadata.lab_id` under this `direct_ancestor` section, for when the sample's parent is another sample/organ.

src/schema/provenance_schema.yaml
```diff
@@ -874,6 +880,11 @@ ENTITIES:
     excluded_properties_from_public_response:
       - lab_tissue_sample_id
       - submission_id
```
yuanzhou, line 882: Add `metadata.lab_id`, for when the sample itself contains such info nested under `Sample.metadata`

yuanzhou (changes requested):

src/schema/provenance_schema.yaml
kburke: It seemed so much clearer to me to get the exclusion fields list for each entity type we recursed through.  But since we're flattening it out in YAML like this, I've added comments.

src/schema/provenance_schema.yaml
kburke: For my Collection test data, the Datasets inside `datasets` did not have `metadata`. But I have made the change.

src/schema/provenance_schema.yaml
kburke: done

src/schema/provenance_schema.yaml
kburke: done

yuanzhou (approved):

PR #795, opened 2025-01-24
Verify Dataset <ID> passed to /prov-metadata before auth checks

kburke:
Fix so /prov-metadata endpoint checks the <ID> is for a Dataset entity before checking authorization levels appropriate for access to whatever entity_type <ID> is for.

### DIFF
--- a/src/dev_entity_worker.py
+++ b/src/dev_entity_worker.py
@@ -281,13 +281,12 @@ class EntityWorker:
     dict
         A dictionary containing all the properties the target entity.
     '''
-    def _get_entity_by_id_for_auth_level(self, entity_id:Annotated[str, 32], valid_user_token:Annotated[str, 32]
-                                         , user_info:dict) -> dict:
+    def _get_expanded_entity_for_auth_level(self, entity_dict:dict, valid_user_token:Annotated[str, 32]
+                                            , user_info:dict) -> dict:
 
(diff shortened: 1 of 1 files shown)


src/dev_entity_worker.py
```diff
@@ -544,14 +543,17 @@ def get_expanded_dataset_metadata(self, dataset_id:Annotated[str, 32], valid_use
...
-                                                                     , valid_user_token=valid_user_token
-                                                                     , user_info=user_info)
+        dataset_dict = self._query_target_entity(entity_id=dataset_id)
+
         # Confirm the dataset_id passed in is for a Dataset entity.
-        if expanded_dataset_dict['entity_type'] not in ['Dataset','Publication']:
-            raise entityEx.EntityBadRequestException(   f"Only Dataset provenance metadata can be retrieved."
-                                                        f" An entity of type '{expanded_dataset_dict['entity_type']}'"
-                                                        f" for uuid={expanded_dataset_dict['uuid']} is not supported.")
+        if dataset_dict['entity_type'] not in ['Dataset', 'Publication']:
```
yuanzhou, line 549: Switch to use `schema_manager.entity_type_instanceof()` to be consistent with other places. And if later we'll add a new subclass of Dataset, we won't have to come back.

yuanzhou (changes requested):

yuanzhou (approved):

Issue #777, opened 2025-01-02
Fix invalid Smart API metadata


yuanzhou:
The entity-api Smart API documentation https://smart-api.info/ui/0065e419668f3336a40d1f5ab89c6ba3 complains about the invalid metadata source.

Verify the yaml spec in https://smart-api.info/editor and fix all the validation errors.

<img width="1130" alt="Screenshot 2025-01-02 at 10 55 30 AM" src="https://github.com/user-attachments/assets/e9f38aa3-5905-43e6-9bbb-4e1eb232694f" />

kburke:
Also look at aligning entity YAML between provenance_schema.yaml and and entity-api-spec.yaml, like [Dataset](https://github.com/hubmapconsortium/entity-api/blob/48dca4dba5ec1f4dbd9a677846616f77452945de/src/schema/provenance_schema.yaml#L299) and [Dataset:properties:](https://github.com/hubmapconsortium/entity-api/blob/48dca4dba5ec1f4dbd9a677846616f77452945de/entity-api-spec.yaml#L520)

Stuff in [entity-api-spec.yaml](https://github.com/hubmapconsortium/entity-api/blob/48dca4dba5ec1f4dbd9a677846616f77452945de/entity-api-spec.yaml#L520) "Datasets" but not
provenance_schema.yaml "Datasets":
```
creators
intended_dataset_type
intended_organ
next_revision_uuids
previous_revision_uuids
```
Stuff in [provenance_schema.yaml](https://github.com/hubmapconsortium/entity-api/blob/48dca4dba5ec1f4dbd9a677846616f77452945de/src/schema/provenance_schema.yaml#L299) "Datasets" but not entity-api-spec.yaml "Datasets":
```
assigned_to_group_name
dataset_info
dataset_type
direct_ancestor_uuids
ingest_id
ingest_task
lab_dataset_id
new_associated_multi_assay_uuid
pipeline_message
provider_info
run_id
status_history
superseded_associated_processed_component_uuids
```

PR #799, opened 2025-01-30
Correct Smart API YAML errors and use security schema

kburke:
Smart API YAML revisions for [entity-api#777](https://github.com/hubmapconsortium/entity-api/issues/777). Correct errors reported in editor at smart-api.info. Use security defined in `securitySchemas` section.


yuanzhou (approved):
@kburke Thanks for the fixes and I just tested in the SmartAPI Editor, the auth works and I was able to get back the desired results from `/datasets/<id>/prov-metadata`.

PR #829, opened 2025-03-25
Addition of Upload fields

kburke:
Initial addition of `anticipated_complete_upload_month` and `anticipated_dataset_count` fields for `Upload`.


kburke:
Pushed changes per review requests.

PR #867, opened 2025-05-14
Added new property priority_project_list and accompanying validator

DerekFurstPitt:


src/schema/provenance_schema.yaml
```diff
@@ -659,6 +659,16 @@ ENTITIES:
           - validate_in_admin_group
         before_property_update_validators:
           - validate_in_admin_group
+      priority_project_list:
+        type: list
+        required: false
+        indexted: false
+        description: The list of priority projects that this Dataset data will be used For
+        mutable: true
```
yuanzhou, line 667: I don't think we have this `mutable` property, only `immutable`.

src/schema/provenance_schema.yaml
```diff
@@ -659,6 +659,16 @@ ENTITIES:
           - validate_in_admin_group
         before_property_update_validators:
           - validate_in_admin_group
+      priority_project_list:
```
yuanzhou, line 662: Because Publication inherits Dataset properties, we'll need to explicitly specify this new field and mark it with `null` to exclude from Publication.

yuanzhou (changes requested):

src/schema/provenance_schema.yaml
DerekFurstPitt: Good catch. I was copying from the details in the card and forgot its immutable not mutable. Fixed this and reversed them (immutable = false rather than mutable = true)

src/schema/provenance_schema.yaml
DerekFurstPitt: Yep, good catch. Fixed

yuanzhou (approved):

PR #964, opened 2025-11-20
Derek furst/validate direct ancestors

DerekFurstPitt:
added a new validator to check whether dataset direct ancestors are only datasets, publications, and samples. Organ samples not allowed


src/schema/schema_validators.py
```diff
@@ -11,6 +11,7 @@
 from schema import schema_neo4j_queries
 from schema.schema_constants import SchemaConstants
 from hubmap_commons import hm_auth
+from app_neo4j_queries import get_entities_by_uuid
```
yuanzhou: @DerekFurstPitt Do we still need this import with your new implementation?

PR #996, opened 2026-03-25
Derek furst/disable s3

yuanzhou:
Following merged PR #992 for TEST


yuanzhou:
Nginx and uWSGI config updates to address #989

PR #997, opened 2026-03-25
Update docker-compose.yml

yuanzhou:


yuanzhou:
Part of the solution to address issue #989

yuanzhou:
Nignx and uWSGI tweaks to address issue #989

PR #1017, opened 2026-06-05
Derek furst/retracted status

DerekFurstPitt:


yuanzhou (changes requested):
Since we've decided to remove the `sub_status` field, it should also be removed from `entity-api-spec.yaml`. And also remove the validator `validate_retracted_dataset_sub_status_value()`.

PR #1022, opened 2026-07-01
Derek furst/doi forwarding

DerekFurstPitt:


yuanzhou:
PR looks good, converted to draft as we are in the process of finalizing the DOI details. Two tweaks will be needed when we are ready to merge:

- Update the path `refs/heads/main/doi-redirect` as we won't need more than one layers
- Add `SECONDARY_REDIRECTION_INFO_URL` to app config example file

yuanzhou (approved):

yuanzhou (approved):

